Issue #174, opened 2014-04-22
Default datatype to plainLiteral


ignazio1977:
Comment by Peter Ansell:
This will break on the update to RDF-1.1, where the datatype is always populated, but not in the way the OWL WG suggested. This will occur when using Sesame-2.8 which contains this update. Instead of the OWL WG path of using RDF_PLAIN_LITERAL for both plain literals and language-tagged literals, the RDF-1.1 group went for two different datatypes, xsd:string and rdf:LangString, for the two subsets of the rdf:PlainLiteral space, respectively.

If, in OWLAPI, both of these datatypes are to be aggregated under rdf:PlainLiteral, then there needs to be a check here to convert both xsd:string and rdf:LangString to rdf:PlainLiteral. Alternatively, xsd:string could be left as is, and rdf:PlainLiteral would only practically be used for language-tagged literals.

ignazio1977:
I believe we've gone with the proper solution a while back

Issue #172, opened 2014-04-22
OWL 2 is not compatible with RDF 1.1 


sesuncedu:
Peter @ansell points out in a comment on  #165 that RDF 1.1 changes the interpretation of literals  without explicit datatypes.

If the literal has no explicit datatype, and no language tag, instead of being a rdf:PlainLiteral with   no language, it is now an xsd:string.

If  the literal has no explicit datatype, but does have a language tag, it is now an rdf:langString. 

The [OWL2 RDF mapping](http://www.w3.org/TR/2012/REC-owl2-mapping-to-rdf-20121211/)  document  continues to state in Table 1 that

```
abc@"^^rdf:PlainLiteral ===>            "abc"
"abc@langTag"^^rdf:PlainLiteral ===>    "abc"@langTag
```

and 
The AS/FSS document states 

> Literals represent data values such as particular strings or integers. They are analogous to typed RDF literals [RDF Concepts] and can also be understood as individuals denoting data values. Each literal consists of a lexical form, which is a  string, and a datatype; the datatypes supported in OWL 2 are described in more detail in Section 4. A literal consisting of a lexical form "abc" and a datatype identified by the IRI datatypeIRI is written as "abc"^^datatypeIRI. Furthermore, literals whose datatype  is rdf:PlainLiteral can be abbreviated in functional-style syntax ontology documents as plain RDF literals [RDF Concepts]. These abbreviations are purely syntactic shortcuts and are thus not reflected in the structural specification of OWL 2. The observable behavior of OWL 2 implementation must be as if these shortcuts were expanded during parsing.

Literals of the form "abc@"^^rdf:PlainLiteral should be abbreviated in functional-style syntax ontology documents to "abc" whenever possible.
Literals of the form "abc@langTag"^^rdf:PlainLiteral where "langTag" is not empty should be abbreviated in functional-style syntax documents to "abc"@langTag whenever possible.

the rdf:PlainLiteral specification adds:

> To eliminate another source of syntactic redundancy and to retain a large degree of interoperability with applications that do not understand the rdf:PlainLiteral datatype, the form of rdf:PlainLiteral literals in syntaxes for RDF graphs and for SPARQL is the already existing syntax for the corresponding plain literal, not the syntax for a typed literal. Therefore, typed literals with rdf:PlainLiteral as the datatype are considered by this specification to be not valid in syntaxes for RDF graphs or SPARQL.

TL;DR - Doooooooooom!.

PR #218, opened 2014-06-30
RDF-1.1 changes

ansell:
This pull request makes the necessary changes to deprecate rdf:PlainLiteral and replace it with the combination of rdf:langString and xsd:string to match RDF-1.1.

This should be taken in the context of the discussion at various issues, including #172 and #174


ansell:
2.8 has been released. Given that version 4 has been released it may be too late to migrate to RDF-1.1. Should I retarget this at version 5 or close it?

ansell:
Closing this to refocus on Sesame-4 and OWLAPI-5

PR #342, opened 2015-02-11
use custom trove hash strategy

sesuncedu:
Use a custom trove hashing strategy in multimap.  

This strategy uses == as a shortcut for true, and hashCode != hashCode as a shortcut for false.  This avoid having to avoids having to make changes to all the axiom implementations to get a win.

(There are also a few stray reformatting related oopses from the obo javadoc fix and an accidental sufficiently large value of 2)


ignazio1977:
merged

PR #350, opened 2015-02-18
Add legacy files  to the OSGI distribution for  version 3.5 backwards compatibility

sesuncedu:
Add  classes org.coode.owlapi.obo.renderer.OBOFlatFileOntologyStorer and  uk.ac.manchester.cs.owl.explanation.ordering.DefaultExplanationOrderer to satisfy legacy Protegé dependencies. 

The stub source is in osgidistribution (to avoid a cyclic dependency on owlapibinding). 

This is not a good location for real code.

Fortunately this is not real code.


sesuncedu:
Any objections?

ignazio1977:
Not from me. Can't wait to drop this unreal code for good :-)

sesuncedu:
T that

PR #353, opened 2015-02-26
Simplify trivially transitive check

tsarkov:
Property is trivially transitive if it can not participate in chains of length more than 1, i.e. s0 -R-> s1 -R-> s2 is inconsistent for (anonymous) s0, s1, s2. This check is used in class InferredObjectPropertyCharacteristicAxiomGenerator to mark trivially transitive property as non-transitive. While questionable on its own (the property **is** transitive in the end), the implementation is really complicated, and involve checking entailment of a set of 2 axioms with anonymous individuals. The proposed implementation is much simpler and gives the same results in almost all cases (the results will only differ on the models with domains of size 1 or 2, which is rather artificial situation). It should also work with any reasoner that allows satisfiability checking.

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/util/InferredObjectPropertyCharacteristicAxiomGenerator.java
+++ b/api/src/main/java/org/semanticweb/owlapi/util/InferredObjectPropertyCharacteristicAxiomGenerator.java
@@ -12,18 +12,14 @@
  * Unless required by applicable law or agreed to in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under the License. */
 package org.semanticweb.owlapi.util;
 
-import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Set;
 
 import javax.annotation.Nonnull;
 
-import org.semanticweb.owlapi.model.AxiomType;
-import org.semanticweb.owlapi.model.OWLAnonymousIndividual;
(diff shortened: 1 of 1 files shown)


sesuncedu:
This version is better.  I would not describe the previous  implementation as too complicated; I would describe it as too simple  (it's a naive translation of FOPL).
 
The filter is needed because of the unfortunate consequences of marking a trivially  transitive property as being transitive:
Given:  
```
Class: A
    DisjointWith: 
        B

Class: B

ObjectProperty: property
    Domain: 
        A   
    Range: 
        B    
```
 The InferredObjectPropertyCharacteristicAxiomGenerator would generate: 
```
ObjectProperty: property
    Characteristics: 
        Irreflexive,
        Asymmetric,
        Transitive
```

Adding Transitive makes  ```property``` [non-simple](http://www.w3.org/TR/2012/REC-owl2-syntax-20121211/#Global_Restrictions_on_Axioms_in_OWL_2_DL) ; that does not go well with Asymmetric:
```
org.semanticweb.owlapi.reasoner.OWLReasonerRuntimeException: Non-simple object property 'http://www.semanticweb.org/ses/ontologies/2015/1/untitled-ontology-24#property' is used as a simple one
	at uk.ac.manchester.cs.factplusplus.FaCTPlusPlus.realise(Native Method)
```

tsarkov:
I see now. Indeed, the transitivity restriction taken blindly is too coarse-grained for the simple role definition. Sigh.

Ignazio, could you please apply this patch to all branches (I'm interested in fixing this for OWLAPI 3.5 and Protege), or shall I produce more pull requests for this?

ignazio1977:
Yes I will.

PR #355, opened 2015-02-27
Support preserving, transferring, and adding annotations during OBO Macro expansion

sesuncedu:
The OBO macro expander was discarding all axiom annotations, even if an axiom didn't have any macros anywhere near it. 

This change set allows annotations to be preserved, transferred, and added to to. 
The changes also allow the macro expander return sets of axioms to add or remove, rather than always add.


cmungall:
+1 to the PR based on quick manual inspection of the code

(does travis build bundled jars for forks? would help lazy sods like me)

ignazio1977:
@cmungall  it should but I've never tried accessing the build output

PR #357, opened 2015-03-01
Support separation of unit and integration tests using junit categories.

sesuncedu:
This change set 
1. Adds a new project, owlapi-testsupport, to hold the required marker interface.   This project does not have owlapi-parent as parent.
2. Adds owlapi-testsupport as a test dependency in owlapi-parent.
3. Excludes the category from surefire config in owlapi-parent.
4. Includes the category in failsafe config in owlapi-parent. 

The change set also makes VerifyVersionInfoTestCase an integration test, and sets Implementation-Version in the bundle instructions so that it passes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: java
+script: "mvn verify"
 jdk:
   - openjdk7
   - oraclejdk7
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+cache:
+  directories:
+    - $HOME/.m2
+
 language: java
 script: "mvn verify"
 jdk:
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -74,11 +74,11 @@
 				<executions>
 					<execution>
 						<id>attach-javadoc</id>
-						<phase>install</phase>
 						<goals>
 							<goal>jar</goal>
 						</goals>
 						<configuration>
+							<skip>${no-javadoc}</skip>
 							<includeDependencySources>true</includeDependencySources>
 							<dependencySourceIncludes>
 								<dependencySourceInclude>net.sourceforge.owlapi:owlapi-api</dependencySourceInclude>
@@ -105,7 +105,6 @@
 				<executions>
 					<execution>
 						<id>attach-sources</id>
-						<phase>install</phase>
 						<goals>
 							<goal>single</goal>
 						</goals>
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -66,15 +66,6 @@
 							org.obolibrary.*
 						</Export-Package>
 					</instructions>
-					<executions>
-						<execution>
-							<id>bundle-manifest</id>
-							<phase>install</phase>
-							<goals>
-								<goal>manifest</goal>
-							</goals>
-						</execution>
-					</executions>
 				</configuration>
 			</plugin>
 			<plugin>
@@ -83,6 +74,7 @@
 				<executions>
 					<execution>
 						<id>attach-javadoc</id>
+						<phase>install</phase>
 						<goals>
 							<goal>jar</goal>
 						</goals>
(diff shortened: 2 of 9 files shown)


sesuncedu:
Any comments?

ansell:
It looks good.

Having VerifyVersionInfoTestCase as an integration test is definitely a good thing, as it requires the full packaging to reliably succeed, and is not something we need to regularly unit test for short-term regressions.

.travis.yml
```diff
@@ -1,5 +1,10 @@
+cache:
+  directories:
+    - $HOME/.m2
```
ansell: So the theory behind starting to do this is that we are going to only be doing mvn verify below and that should ensure that build artifacts from previous builds never get cached?

distribution/pom.xml
```diff
@@ -83,6 +74,7 @@
 				<executions>
 					<execution>
 						<id>attach-javadoc</id>
+						<phase>install</phase>
```
ansell: If these plugins are painful (slowness is the only reason I could think of for delaying their creation), we could create a specific profile for them, rather than assembling them in the same phase that they are being pushed to the local repository in. The order of plugins in the install phase would then determine exactly how the artifacts were pushed, which doesn't seem like it would be a good thing in the long term if builds are not repeatable.

testsupport/pom.xml
```diff
@@ -0,0 +1,99 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <groupId>net.sourceforge.owlapi</groupId>
+    <version>3.5.2-SNAPSHOT</version>
+    <modelVersion>4.0.0</modelVersion>
```
ansell: My personal preference is to put modelVersion at the top so that it isn't confused for version.

distribution/pom.xml
sesuncedu: I'm not sure if this forces the artifact to be built or not... MVN is  This is a job for **-X**

ansell: The reasoning I thought for doing it would be to enable mvn verify to run without having these built, by pushing them into the next phase in the pipeline, install. The ordering of plugins within a phase is only done by the order in the pom.xml files that are used, so it may not be what is expected in general from Maven.

I am not quite sure what you mean by forcing the artifact to be build or not.

Caching the .m2 directory is quite difficult in general for maven, as it doesn't distinguish the paths for locally installed packages from remote repository fetched packages. It only distinguishes them inside of the metadata files that it keeps.

It would be useful to restrict this pull request to the junit categories change, and split out the travis caching changes into another pull request, as the categories change looks fine on its own but there are some ramifications to changing the plugin phases that aren't always clear up front.

sesuncedu: And it will consistently... not work (except for the release profile, which is untouched). I will take a brief look to see if there is any way to force the attach to happen before the install plugin ; otherwise I'll go with the profile option.   If only the javadoc plugin was incremental.

ansell: I am fine with switching the javadoc plugin off for normal compiles, and only having it active in the release profile. If users want to generate the javadoc out of the git without releasing, they can manually run mvn javadoc:jar

sesuncedu: New commit defines a user property ${no-javadoc} which defaults to true, and uses that to set maven.javdoc.skip .  

To generate javadoc use mvn package -Dno-javadoc=false

sesuncedu: The caching was supposed to be for dependencies.

http://docs.travis-ci.com/user/caching/

It turns out there needs to be one more line in the travis config (sudo:
false) to make it use the container based infrastructure.

http://docs.travis-ci.com/user/workers/container-based-infrastructure/

I'm doing stuff at Protégé that needs fixes in 3.5.2, so I am doing a lot
of dependency fun and games with locally installed snapshots (including
discovering that reasoner versions can't have non digits. And that you
can't have a SNAPSHOT of a beta-16 and expect maven to work.

ignazio1977:
Merged

PR #358, opened 2015-03-02
Solve some thread safety issues in XML parsers and storers

sesuncedu:
Use thread locals instead of static variables (issues don't show up in tests until multi-threading is enabled. Then they show up).

### DIFF
--- a/parsers/src/main/java/org/coode/xml/XMLWriterImpl.java
+++ b/parsers/src/main/java/org/coode/xml/XMLWriterImpl.java
@@ -64,6 +64,7 @@ import org.semanticweb.owlapi.vocab.OWL2Datatype;
  */
 public class XMLWriterImpl implements XMLWriter {
 
+    private final XMLWriterPreferences instance = XMLWriterPreferences.getInstance();
     private Stack<XMLElement> elementStack;
     protected Writer writer;
     private String encoding = "";
@@ -305,7 +306,7 @@ public class XMLWriterImpl implements XMLWriter {
             encodingString = " encoding=\"" + encoding + "\"";
         }
         writer.write("<?xml version=\"1.0\"" + encodingString + "?>\n");
-        if (XMLWriterPreferences.getInstance().isUseNamespaceEntities()) {
+        if (instance.isUseNamespaceEntities()) {
             writeEntities(rootElement);
         }
         preambleWritten = true;
@@ -494,7 +495,7 @@ public class XMLWriterImpl implements XMLWriter {
             writer.write(attr);
             writer.write('=');
             writer.write('"');
-            if (XMLWriterPreferences.getInstance().isUseNamespaceEntities()) {
+            if (instance.isUseNamespaceEntities()) {
                 writer.write(swapForEntity(XMLUtils.escapeXML(val)));
             } else {
                 writer.write(XMLUtils.escapeXML(val));
@@ -525,9 +526,9 @@ public class XMLWriterImpl implements XMLWriter {
         }
 
         private void insertIndentation() throws IOException {
-            if (XMLWriterPreferences.getInstance().isIndenting()) {
+            if (instance.isIndenting()) {
                 for (int i = 0; i < indentation
-                        * XMLWriterPreferences.getInstance().getIndentSize(); i++) {
+                        * instance.getIndentSize(); i++) {
                     writer.write(' ');
                 }
             }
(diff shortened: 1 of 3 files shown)


parsers/src/main/java/org/semanticweb/owlapi/rdf/syntax/RDFParser.java
```diff
@@ -84,29 +84,45 @@
...
+
+    protected static Locator getLocator() {
+        return locatorThreadLocal.get();
+    }
+    private static ThreadLocal<SAXParserFactory> saxParserFactoryThreadLocal = new ThreadLocal<SAXParserFactory>() {
+
+        @Override
+        protected SAXParserFactory initialValue() {
+            try {
+                System.setProperty("entityExpansionLimit", "100000000");
```
ansell: Can this be called once in a static context to avoid setting it for each thread?

parsers/src/main/java/org/coode/xml/XMLWriterImpl.java
```diff
@@ -64,6 +64,7 @@
  */
 public class XMLWriterImpl implements XMLWriter {
 
+    private final XMLWriterPreferences instance = XMLWriterPreferences.getInstance();
```
ansell: So this class is never shared between threads?

parsers/src/main/java/org/semanticweb/owlapi/rdf/syntax/RDFParser.java
sesuncedu: SAXParserFactory itself is not thread safe.

ansell: I just meant the System.setProperty call that this comment is against, not the lines below this one. Not being familiar with System.setProperty, I would think that you wouldn't need to call it multiple times for a given JVM (assuming noone is maliciously or accidentally changing it).

parsers/src/main/java/org/coode/xml/XMLWriterImpl.java
sesuncedu: I don't think so. Given how things blew up when one thread changed the use entities setting whilst another class was writing its output (thus causing it to assume that it had defined entities when starting the document, and thus start using entities that were in fact undefined). 

  The reason for extracting  to a field is that the preference object is accessed for every line written to check for indentation settings.  
The XML formats have a lot of lines.

parsers/src/main/java/org/semanticweb/owlapi/rdf/syntax/RDFParser.java
sesuncedu: The non-thread-safe XML factory/builder classes drive me nuts.

ignazio1977: We had troubles with threadlocals in Tomcat a long time back - if I remember correctly, around the caches for IRI prefixes and other similar caches. I don't think the creation of a SaxParserFactory happens often enough or is expensive enough to bother caching it - we could do this the same way it's done in version 5

https://github.com/owlcs/owlapi/blob/version5/api/src/main/java/org/semanticweb/owlapi/util/SAXParsers.java

It's not perfect because RDFParser still has a static reference, so that needs addressing as well.
Good work spotting this, by the way.

sesuncedu: There's another thread safety issue somewhere in either the obo parser or
the OWL XML parser ; it shows up once the RDF / XML race is stopped,
manifesting as some warnings that some OWL/XML string can't be parsed by
OBO. However there is no hard failure to cause the unit tests to fail, so
it's more work to catch.

It's most likely an OBO issue.  I will be poking in the OBO stuff later; if
I see something I will fix it, but I won't be looking too hard right now.

The thread fun is trying to make Protégé not do *everything* on the EDT,
which is less than fun when working with GO :)

ignazio1977: I've made a different fix in version 4, without the need to set system properties - the values can be set directly on the generated parser. There is a utility method on SaxParsers to centralize this, so that client code does not need to worry.

sesuncedu: I copied the line from the OWLXML Parser code, in case it's still in
version4?

ignazio1977: No it was refactored in SaxParsers. There should be no setting of system properties any more, I believe - I'll double check just to be safe.

PR #360, opened 2015-03-03
Don't shade everything 

sesuncedu:
(Plus some benchmarks stuff that somehow got mixed in)


sesuncedu:
Note that the shading only happens when the release profile is used. 

Also, the test may fail due to the entity expansion setting change

ignazio1977:
Merged

PR #362, opened 2015-03-07
OSGI Integration Test, and fixes so it can pass

sesuncedu:
There are certain things even John Austin couldn't do with words.  
One of them is express quite how much I hate OSGI right now.  

Actual info in commit messages and diffs.  Except for the missing 'not' in 'not as keen'.


sesuncedu:
There was a Roundtripping test failure on the last commit that went away whilst I looked a travis ; I think there's probably race conditions still.

Also: 
https://github.com/ops4j/peaberry/wiki/Peaberry-Activation-Guide

sesuncedu:
Tests in error: 
  PrimerOWLXMLRoundTrippingTestCase>AbstractRoundTrippingTestCase.testManchesterOWLSyntax:80->TestBase.roundTripOntology:440->TestBase.equal:140 » IllegalArgument

ignazio1977:
You and me both. I was trying to do a release yesterday but the way the testsupport module is wired in is not liked by Nexus. Ah well.

ignazio1977:
I've merged this. No exceptions locally, we'll see what Travis says

Issue #363, opened 2015-03-08
Major bug in punning detection, concealing major performance problem in filterAxioms


sesuncedu:
Whilst looking in to the issue of why some parts of ontology rendering were non-deterministic, I discovered a major problem in the [faster punning detection code](https://github.com/owlcs/owlapi/blob/version4/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLImmutableOntologyImpl.java#L609)

```java
public Set<IRI> getPunnedIRIs(Imports includeImportsClosure) {
        Set<IRI> punned = new HashSet<>();
        Set<IRI> test = new HashSet<>();
  [includeImportsClosure case omitted]
       for (OWLEntity e : getClassesInSignature(EXCLUDED)) {
            punned.add(e.getIRI());
        }
        for (OWLEntity e : getDataPropertiesInSignature(EXCLUDED)) {
            if (test.add(e.getIRI())) {
                punned.add(e.getIRI());
            }
        }
        for (OWLEntity e : getObjectPropertiesInSignature(EXCLUDED)) {
            if (test.add(e.getIRI())) {
                punned.add(e.getIRI());
            }
        }
        [etc.]
        return punned;
```
This defines defines a punned IRI as one that names a class, a data property, an object property that is not also a data property, etc. 

In some ways this is a good thing, as correcting the error causes  saving the Gene Ontology as RDF/XML to take half an hour, almost all of the time being in MapPointer.filterAxioms. 

Investigation continues. 

Also, Annotation Assertions are not sorted. 
Also, predicates in triples are not totally ordered.

sesuncedu:
Ah...  this would be the filter Axioms by doing a linear search through every single annotation property assertion axiom instead of using the index approach.

sesuncedu:
Internals::filterAxioms is about to get a minimalist hack.

sesuncedu:
After, 3.5s. That's a bit better.  Now to see what broke

ignazio1977:
Rings a bell, I think I fixed something similar a month or two ago. Core Developer Misusing Own API Badge Of Shame: ASSIGNED :-(

ignazio1977:
I've changed a number of calls to Filters.annotations that I had missed first time around. Really that filter should go.
I won't touch the area at the moment, if you're making changes as well.

sesuncedu:
Change made.  I didn't pull request because I  was feeling too tired, and I
hit a weird writer bug in a non test suite run against some modded
[...]

PR #364, opened 2015-03-08
Code to generate more stable renderings.

sesuncedu:
Fix PunnedIRI problem 
Sort more axioms. 
Sort more triples.

Make rdf predicates totally ordered. 

Special case filter axioms for annotations (mootish but not yet worthless).
Addresses some issues in  https://github.com/owlcs/owlapi/issues/363


ignazio1977:
Merged

impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/Internals.java
```diff
@@ -877,8 +878,18 @@ public boolean isEmpty() {
      * @return set of values
      */
     @Nonnull
-    public <K> Collection<OWLAxiom> filterAxioms(
+    public <K> Collection<? extends OWLAxiom> filterAxioms(
             @Nonnull OWLAxiomSearchFilter filter, @Nonnull K key) {
+        if(filter == Filters.annotations) {
```
ansell: This would be a little easier to parse if the constants in Filters were in upper-case, ie, Filters.ANNOTATIONS, but I guess changing that in version 4 is probably out of the question now.

PR #369, opened 2015-03-11
Manimal  - jdk 1.6 compatibility fixes, with maven animal-sniffing...

sesuncedu:
Do not upgrade animal sniffer to 1.14 as it will throw an Null Pointer Exception.

I'm not 100% sure if the comparison code is right WRT OWL/RDF/XSD, but I don't think it's supposed to be. I just care about literals sorting consistently, which is all these methods are for, I believe. Proper ordering by datatype is a job for the reasoner. SEP field engaged.

### DIFF
--- a/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplBoolean.java
+++ b/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplBoolean.java
@@ -195,6 +195,11 @@ public class OWLLiteralImplBoolean extends OWLObjectImpl implements OWLLiteral {
 
     @Override
     protected int compareObjectOfSameType(OWLObject object) {
+        if (object instanceof OWLLiteralImplBoolean) {
+            OWLLiteralImplBoolean aBoolean = (OWLLiteralImplBoolean) object;
+            boolean other = aBoolean.literal;
+            return compareBoolean(literal, other);
+        }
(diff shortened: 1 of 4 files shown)


impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplBoolean.java
```diff
@@ -204,7 +209,21 @@ protected int compareObjectOfSameType(OWLObject object) {
...
+
+    private static int compareBoolean(boolean a, boolean b) {
+        if (a) {
+            if (b) {
+                return 0;
+            } else {
+                return 1;
+            }
+        } else if (!b) {
+            return 0;
```
ansell: This case looks a little strange. If a and b are reversed it will return -1 from the case below, which doesn't fulfill the general Comparable contract.

In the past I have used something like the following from memory:

    if (a!=b) {
        if (b) {
            return 1;
        } else {
            return -1;
    } else {
        return 0;
    }

ignazio1977: Patch set updated accordingly and merged

ignazio1977:
merged separately

ignazio1977:
Good stuff, I'm going to create signatures for the previous versions of OWLAPI and use them locally for testing (don't fancy releasing the signature artifacts at this time)

PR #370, opened 2015-03-12
Sorting of FSS 

sesuncedu:
Render FSS with more sorting. 
Generate comments for sections and entities. 

Only reason render a section and entities if there are non declarations to be handled. 

To do ; GCI section header 
Refactoring to common base for other renderings.


ignazio1977:
This looks good. I'm starting to wonder if all the sorting that's going on isn't a sign that we should keep this stuff in sorted collections (or, for immutable objects, lists or arrays sorted on parsing/building).

PR #376, opened 2015-03-16
Simple OWL/FSS ontology IRI sniffing for AUTO IRI mapper. 

sesuncedu:


ignazio1977:
Good stuff, I think there was a feature request for this

PR #379, opened 2015-03-23
Support OWLAnnotation caching to data factory 

sesuncedu:
Add  OWLAnnotation object creation method to Data Factory Internals.
Use a guava  LoadingCache to provide a size-limited pseudo-interner  for OWLAnnotations.

This change can be a win for OBO family ontologies that use some literal valued axiom annotations repeatedly. 

It cuts about 10% off the size off the size of GO. It is relatively less effective for NCBI taxon, despite having an almost perfect hit rate.  This is because the values for those annotations are IRIs, which are already cached. 

Looking at this code reminds me that it's about time to switch the caches to guava, and use delegation rather than inheritance for caching.  This fix is for a specific use case; The next priority  is to  use singleton sets / small sets for annotations where appropriate.


ignazio1977:
merged

PR #446, opened 2015-10-26
Integrate with Commons RDF API for RDF concepts

stain:
[Apache Commons RDF](http://commonsrdf.incubator.apache.org/) (incubating) is an effort for making a common API (e.g. Java interfaces) for RDF concepts like [`IRI`](http://commonsrdf.incubator.apache.org/apidocs/org/apache/commons/rdf/api/IRI.html) and [`BlankNode`](http://commonsrdf.incubator.apache.org/apidocs/org/apache/commons/rdf/api/Literal.html). Commons RDF aims to become a common interface between frameworks like Jena and Sesame so that you can have direct API compatibility without having to serialize and deserialize RDF. 

## Motivation

* Commons RDF tries to stay as close as possible to the [RDF 1.1 concepts](http://www.w3.org/TR/rdf11-concepts/) and to be neutral common ground between the Java RDF frameworks.
* The Commons RDF API is now stabilizing, although we appreciate any [feedback](http://commonsrdf.incubator.apache.org/contributing.html) or suggested modifications before releasing 1.0.0.
* Commons RDF integration is planned in most of the Java framework projects, e.g. Apache Jena [JENA-1015](https://issues.apache.org/jira/browse/JENA-1015), Sesame [SES-2091](https://openrdf.atlassian.net/browse/SES-2091) and Apache Clerezza.
* Integrating Commons RDF with OWLAPI means that users (and OWLAPI) has the potential to combine the two in a more direct way, and can use parsers, writers and triple stores from these frameworks.   

## Implementation

This patch shows the OWLAPI modifications needed to the types `RDFLiteral`, `RDFResource`, `RDFNode`, `RDFLiteral`, `RDFTriple` and `IRI`.  Changes are generally minimal, as the overlap was already big.

OWLAPI doesn't really have the concept of a `Graph` as statements are generated or consumed on the fly.

OWLAPI has both `IRI` and `RDFResourceIRI` - I am not sure why both are needed, but it was easy enough to make both of them compatible with Common RDF's `IRI`.

Note that Commons RDF also expects [.equals()](http://commonsrdf.incubator.apache.org/apidocs/org/apache/commons/rdf/api/IRI.html#equals-java.lang.Object-) to work across implementations on the core concepts, so therefore this patch also modifies OWLAPI .equals() to support a fallback to other Common RDF instances. 

I included a `CommonsRDFTermTest` to use Common RDF test suite. Some of the failing tests are ignored as they deal with validation of things like the Literal language string, which I am not sure OWLAPI needs to do as it is not an RDF framework.

## Further work

Further work would be needed on this branch to support RDF framework integration smoothly, e.g. as a special kind of OWLAPI `RDFConsumer` or `RDFRendererBase` that takes a Commons RDF [`Graph`](http://commonsrdf.incubator.apache.org/apidocs/org/apache/commons/rdf/api/Graph.html) instance instead of a file and format.

This integration could be taken further by removing some or all of the RDF* types in favour of the Commons RDF Simple implementation, like where there is no additional benefits given - however this would mean adding API changes from OWLAPI 4.

This patch highlights a few potential method renames in the OWLAPI to be directly compatible and avoid duplicate methods, however I have not suggested that as part of this patch as it is more drastic:

* RDFLiteral.getLexicalValue() -> getLexicalValue()
* RDFLiteral.getLang() -> getLanguageTag()
* IRI.toQuotedString() -> ntriplesString()

I also found that `RDFLiteral` deals with RDF 1.0 strings and uses `rdf:PlainLiteral` (RDF 1.1 merges plain literals to `xsd:string`) - I've tried to keep `rdf:PlainLiteral` support, but output both of those without `^^` type in the new `literal.ntriplesString()`, as Commons RDF follows RDF 1.1 specifications.  I also made the `rdf:langString` as a default type if a lang is given - but didn't force this as a type if BOTH lang and data type is given. (RDF 1.1 specs however says it must be `rdf:langString in this case) --- should this be raised as separate bugs?

There could be some potential bugs in OWLAPI highlighted in terms of equivalence, as OWLAPI BlankNodes can carry with them their local identifier, but that means that two such blank nodes from two different parser runs could potentially accidentally overlap if they were both added to a common Commons RDF `Graph` instance.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -14,5 +14,10 @@
 			<artifactId>xz</artifactId>
 			<version>1.5</version>
 		</dependency>
+	  <dependency>
+	      <groupId>org.apache.commons</groupId>
+	      <artifactId>commons-rdf-api</artifactId>
+	      <version>0.1.0-incubating</version>
+	  </dependency>
 	</dependencies>
 </project>
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -14,10 +14,17 @@
 			<artifactId>xz</artifactId>
 			<version>1.5</version>
 		</dependency>
-	  <dependency>
-	      <groupId>org.apache.commons</groupId>
-	      <artifactId>commons-rdf-api</artifactId>
-	      <version>0.1.0-incubating</version>
-	  </dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-rdf-api</artifactId>
+			<version>0.1.0-incubating</version>
+		</dependency>
+		<dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-rdf-api</artifactId>
+			<version>0.1.0-incubating</version>
+			<classifier>tests</classifier>
+			<scope>test</scope>
+		</dependency>	  
 	</dependencies>
 </project>
--- a/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
+++ b/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
@@ -188,7 +188,8 @@ public class RDFLiteral extends RDFNode implements org.apache.commons.rdf.api.Li
   		String escaped = '"' +
   				EscapeUtils.escapeString(getLexicalValue()).
   				replace("\n", "\\n").replace("\r", "\\r") + '"';
-  		if (datatype.equals(OWL2Datatype.RDF_PLAIN_LITERAL.getIRI())) {
+  		if (datatype.equals(OWL2Datatype.RDF_PLAIN_LITERAL.getIRI()) || 
+  			datatype.equals(OWL2Datatype.XSD_STRING.getIRI())) {
   			return escaped;
   		} else if (hasLang()) {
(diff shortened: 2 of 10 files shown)


stain:
See also #218 for RDF 1.1 integration.

I get an Travis error: 
> Failed tests: 
>   RioUtilsTestCase.testTripleLangLiteral:146 expected:<(urn:test:subject:uri:1, urn:test:predicate:uri:1, "Test literal"@en)> but was:<(urn:test:subject:uri:1, urn:test:predicate:uri:1, "Test literal"^^<http://www.w3.org/1999/02/22-rdf-syntax-ns#langString>)>

Is this related to my patch setting the default type to OWL2Datatype.RDF_LANG_STRING for strings with lang provided? Perhaps some code was mistakenly testing for PlainLiteral type (but strings with `@language` are not plain literals)

ignazio1977:
The problem was checking for null datatype (it used to be the case for plain literals, but that changed a while back) instead of using isPlainLiteral().
Fixed and merged.
I'll put the comment in an issue to remember what's still to be done - great patch.

PR #459, opened 2015-11-12
Initial version of hack - always preserve rdfs:class if present.

sesuncedu:
**DO NOT MERGE**
Test cases failing due to class cast error 
Control via option not done.

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/io/RDFParserMetaData.java
+++ b/api/src/main/java/org/semanticweb/owlapi/io/RDFParserMetaData.java
@@ -39,21 +39,22 @@ public class RDFParserMetaData implements OWLOntologyLoaderMetaData,
     private final RDFOntologyHeaderStatus headerStatus;
     private final Set<RDFTriple> unparsedTriples;
     private final ArrayListMultimap<IRI, Class<?>> guessedDeclarations;
-
+    private final Set<IRI> rdfsClassDeclarations;
     /**
      * @param headerStatus
      *        the header status
      * @param tripleCount
      *        the triple count
      * @param unparsedTriples
-     *        the set of triples not parsed
+ *        the set of triples not parsed
      * @param guessedDeclarations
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/semanticweb/owlapi/io/RDFParserMetaData.java
```diff
@@ -39,21 +39,22 @@
...
+ *        the set of triples not parsed
      * @param guessedDeclarations
-     *        guessed declarations map
+     * @param rdfsClassDeclarations
      */
     public RDFParserMetaData(@Nonnull RDFOntologyHeaderStatus headerStatus,
-            int tripleCount, @Nonnull Set<RDFTriple> unparsedTriples,
-            @Nonnull ArrayListMultimap<IRI, Class<?>> guessedDeclarations) {
+                             int tripleCount, @Nonnull Set<RDFTriple> unparsedTriples,
+                             @Nonnull ArrayListMultimap<IRI, Class<?>> guessedDeclarations, Set<IRI> rdfsClassDeclarations) {
```
ansell: Should we add another version of the constructor that preserves the previous signature and sends an empty set here? That would help preserve backwards compatibility if we have this class as part of our public API.

ansell:
Looks good to me except for the one comment about constructor signatures.

api/src/main/java/org/semanticweb/owlapi/io/RDFParserMetaData.java
sesuncedu: Definitely. Also, I hate having all this stuff in a class that is part of a public API.

sesuncedu:
Needs config settings.
I really wish we could use the rio config classes, but that's too much for
this ugly a hack.

ansell:
Using the Rio config style with named and typed parameters (if not the actual RioConfig class) may be useful to target for version 5.

Issue #512, opened 2016-04-09
SWRL rules with incorrect equals()


Galigator:
Hello,

While converting pellet to owl-api5 (5.0.2-SNAPSHOT git:7561524 )  I have hit the message of the title.

The test case is :
https://github.com/Galigator/pellet/blob/owlapi5/test/src/test/java/com/clarkparsia/explanation/test/AbstractExplanationTest.java
Line : 368

The error doesn't show when using owlapi4

OWL.classAssertion(a, A), 
OWL.propertyAssertion(a, dp, OWL.constant(9)),/
SWRL.rule(
	SWRL.antecedent(
	        SWRL.classAtom(A, x),
		SWRL.propertyAtom(dp, x, dx),
		SWRL.greaterThan(dx, SWRL.constant(5)),
		SWRL.lessThan(dx, SWRL.constant(10))
	),
	SWRL.consequent(SWRL.classAtom(B, x))
) 

Must entail OWL.classAssertion(a, B)

But the following stack trace show :

org.semanticweb.owlapi.model.OWLRuntimeException: Explanation contains removed axiom: DLSafeRule( Body(ClassAtom(<http://www.example.org/test#A> Variable(<http://www.example.org/test#x>)) DataPropertyAtom(<http://www.example.org/test#dp> Variable(<http://www.example.org/test#x>) Variable(<http://www.example.org/test#dx>)) BuiltInAtom(<http://www.w3.org/2003/11/swrlb#lessThan> Variable(<http://www.example.org/test#dx>) "10"^^xsd:integer ) BuiltInAtom(<http://www.w3.org/2003/11/swrlb#greaterThan> Variable(<http://www.example.org/test#dx>) "5"^^xsd:integer )) Head(ClassAtom(<http://www.example.org/test#B> Variable(<http://www.example.org/test#x>))) )
at com.clarkparsia.owlapi.explanation.HSTExplanationGenerator.recurse(HSTExplanationGenerator.java:300)
at com.clarkparsia.owlapi.explanation.HSTExplanationGenerator.constructHittingSetTree(HSTExplanationGenerator.java:242)
at com.clarkparsia.owlapi.explanation.HSTExplanationGenerator.getExplanations(HSTExplanationGenerator.java:120)
at com.clarkparsia.explanation.test.OWLAPIExplanationTest.testExplanations(OWLAPIExplanationTest.java:243)
[...]

PR #514, opened 2016-04-11
Fix equality between two rules. (V2)

Galigator:
Body & Head of rules must be compare using the same ordering.

see #512

This implementations use less details of SWRLRule into OWLObjectImpl.


ignazio1977:
Merged in bdeaa65400489db22f45ca0b3578a5ac9f3c7052

PR #533, opened 2016-06-12
Add addAxiom method for stream support.

Galigator:
The method addAxioms that take a Stream as parameter is missing. Is there a reason ?


ignazio1977:
Nope, no reason, it was just missed.

PR #549, opened 2016-08-04
Change from OpenRDF Sesame to Eclipse RDF4J

ansell:
OpenRDF Sesame is not having any more releases. It has been moved into Eclipse and re-released as Eclipse RDF4J. This pull request makes the necessary changes to support that migration. I am targeting this at 5.0 for peer review, but the changes affect the public API so probably need to be done in OWLAPI-6.

The jackson version is restricted to <2.7 right now due to removals from the Jackson public API in 2.7, but the next RDF4J release should allow 2.7/2.8 to work.

Signed-off-by: Peter Ansell <p_ansell@yahoo.com>


ansell:
The equivalent pull request for Semargl is at:

https://github.com/levkhomich/semargl/pull/45

Until that is merged, Semargl is commented out of the pom.xml files to review the rest of the changes.

ansell:
Retargeting this to master that now has version 6.0.0-SNAPSHOT attached to it

Issue #550, opened 2016-08-06
owlapi parsers should be vetted for leaking console output


cmungall:
Whenever we parse an .obo format file we get an error message on the console:

```
[Fatal Error] :1:1: Content is not allowed in prolog.
```

This is in fact not a obo format error message. The message is coming from a org.xml.sax.SAXParseException that naughtily writes to stderr as well as throwing an exception. the exception is caught and the next parser is tried, but the console output has already leaked.

If an upstream fix can't be made then perhaps the OWLAPI could use some heuristics to choose a more optimal order in which to cycle through it's now rather large list of default registered parsers.

sesuncedu:
+1

This has sat too long for any good ignoring it has done lately.

I think I started chasing this, and got as far as the realization that
there needed to be an errorHandler set, with no easy place to do it quite
the right way.

It should be easy enough to do better than stderr.

ignazio1977:
Files with the .obo extension are a pretty obvious clue to skip XML parsers.

We could also check the first character to skip any XML based parsing, since all sax parsing will fail if the first character is not the start of an XML element (I believe, need to verify that).

ansell:
That particular message may be coming from the Rio RDF/XML (SAX) parser, and I am not sure myself how to stop it as I recall looking into it in the past but couldn't find a solution.

ansell:
What version of OWLAPI and Sesame are you using? It is possible this was fixed in Sesame after the version that your OWLAPI version is bundling as I can't replicate it with a regression test on the RDF4J codebase that is derived from Sesame-4.1.2.  In RDF4J-2.0M3, there is an ErrorHandler setup to pipe the SAX fatal errors through to the RDF4J ParseErrorListener and an RDFParseException is thrown, but System.err and System.out are not written to anymore.

ansell:
This was apparently fixed (by me) in Sesame-2.7.0 so it is strange that it should be appearing again here unless there is another SAXParser in the OWLAPI stack.

Fix references for Sesame are:

https://openrdf.atlassian.net/browse/SES-666

https://bitbucket.org/openrdf/sesame/commits/c1a0bdbc2cc7ae645581b054444bb3f5ab61a55b

ignazio1977:
Nice one Peter.
That's a good fix for unnecessary messages, but the parser prioritisation might have performance advantages as well, so we should probably include both.

sesuncedu:
[...]

PR #553, opened 2016-08-11
Issue #550

sesuncedu:
Fix for Rio Trix parser logging errors to stderr. 

This was so not worth it.

### DIFF
--- a/rio/src/main/java/org/semanticweb/owlapi/rio/utils/OWLAPISimpleSAXParser.java
+++ b/rio/src/main/java/org/semanticweb/owlapi/rio/utils/OWLAPISimpleSAXParser.java
@@ -1,4 +1,8 @@
-/* 
+/*  NOTE:  Derived from openrdf sesame version 2.9.0.  See
+ *  sesame distribution, available at https://bitbucket.org/openrdf/sesame
+ * for license details referenced below.
+ * Cloned from info.aduna.xml.SimpleSaxParser in order to be able to mask errors.
+
  * Licensed to Aduna under one or more contributor license agreements.  
  * See the NOTICE.txt file distributed with this work for additional 
  * information regarding copyright ownership. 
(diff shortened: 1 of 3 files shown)


rio/src/main/java/org/semanticweb/owlapi/rio/utils/OWLAPISimpleSAXParser.java
```diff
@@ -0,0 +1,358 @@
...
+	 * 
+	 * @param inputSource
+	 *        An <tt>InputSource</tt> containing XML data.
+	 */
+	private synchronized void parse(InputSource inputSource)
+		throws SAXException, IOException
+	{
+		SimpleSAXDefaultHandler handler = new SimpleSAXDefaultHandler();
+		xmlReader.setContentHandler(handler);
+		xmlReader.setErrorHandler(handler);
```
ansell: In any other situation I would say this is a bad thing, as DefaultHandler, that SimpleSAXDefaultHandler uses, has noop implementations of the ErrorHandler methods (I had to check that yesterday myself), so it may affect genuine exception handling (don't need to look into that unless people complain)

However, the rabbit hole is very deep and you want to keep the light (sane formats such as TriG for instance) in view so just going with this is good with me.

ansell:
Looks good to me, link for attribution purposes only, and the copyright/license headers are preseved.

ansell:
Part of this pull request (to make sure we don't have to come back to this again!) should also exclude sesame-rio-trix as a dependency, and add the META-INF/services file from sesame-rio-trix here.

sesuncedu:
If I drop sesame-rio-trix and add the META-INF/services stuff then I need to clone the factory, which  means I can drop the special casing in the trix factory.  

Down we go [Do cats eat bats? Do bats eat cats?](https://youtu.be/QdbfMnjH6Ic?t=3m6s)

rio/src/main/java/org/semanticweb/owlapi/rio/utils/OWLAPISimpleSAXParser.java
sesuncedu: The default implementation of [DefaultHandler::fatalError](http://www.saxproject.org/apidoc/org/xml/sax/helpers/DefaultHandler.html#fatalError(org.xml.sax.SAXParseException) rethrows the SaxParseException. It's when this re-thrown error is caught in the TrixParser that the  error listener is invoked (this fooled me at first). 

Also, isn't it great how javax.xml.parsers.SAXParser requires DefaultHandler (as an argument to parse, overriding any handlers set on the underlying parser), but javax.xml.parsers.SAXParser requires individual handlers to be set on the XMLReader.

Issue #571, opened 2016-10-26
Assume object property when property not defined in LAX mode


chrispetsos:
I use the following snippet to load an ontology,
`manager = OWLManager.createOWLOntologyManager();
		try {  
			ontology = manager.loadOntologyFromOntologyDocument(new StreamDocumentSource(axioms));  
		} catch (OWLOntologyCreationException e) {  
			e.printStackTrace();  
		}  
`

ignazio1977:
Can you describe what you wish to happen?

When a property is undefined, parser behavior is supposed to be to throw exceptions.

In order to deal with real world data, we attempt to treat such cases with a best guess, but we need more information on this one.

ignazio1977:
Sorry, saw your update after my comment

chrispetsos:
The new title represents what my original intention was after all... i'll issue a new pull request that tries to fulfill this in a more elegant way.

PR #572, opened 2016-10-27
Enhancement for Issue 571 (https://github.com/owlcs/owlapi/issues/571).

chrispetsos:
This is my first shot regarding issue 571. What do you think?

Relax a bit the Object Property assertion logic. If it is not explicitly
declared to be an Object Property or a Datatype Property, then assume it
is an Object Property.


chrispetsos:
Turns out that this solution to Issue #571 was problematic. Closing this pull request. I've tried another approach in pull request #573.

PR #573, opened 2016-11-03
Enhancement for Issue 571 (https://github.com/owlcs/owlapi/issues/571).

chrispetsos:
When checking for qualified Object or Datatype cardinalities in LAX, it
is enough for the property just to be present, do not check if it is an
actual Object or Datatype property. The rationale behind this relaxation
of checks is that if someone has correctly declared a onDataRange
restriction property, then he/she intends to create a Datatype property
restriction. In an analogous manner, when someone correctly declares a
onClass restriction property then he/she intends to create an Object
property restriction. Note that for non-qualified restrictions the check
has not been relaxed, because we cannot tell what the intention of the
user is.

No need to merge this, but it would be really nice if you could share some thoughts/comments on my approach here. Thanks in advance!

### DIFF
--- a/parsers/src/main/java/org/semanticweb/owlapi/rdf/rdfxml/parser/Translators.java
+++ b/parsers/src/main/java/org/semanticweb/owlapi/rdf/rdfxml/parser/Translators.java
@@ -798,7 +798,7 @@ public class Translators {
         public boolean matchesLax(IRI mainNode) {
             return isNonNegativeIntegerLax(mainNode,
                     OWL_MAX_QUALIFIED_CARDINALITY)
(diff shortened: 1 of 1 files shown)


chrispetsos:
OK Ignazio, yes I understand your concerns. Obviously, my application is benefited from a more relaxed checking policy when it comes to qualified cardinalities and can live with the cases of malformed ontologies around those things. I'll keep on with my own fork... Closing this pull request and thanks for your feedback.

ignazio1977:
I'm happy to have the issue open so that the preference driven switch can be added - otherwise I risk forgetting about it :-)

Issue #574, opened 2016-11-15
Order of RDF triples affects annotation parsing


balhoff:
I've noticed that when parsing ontologies in RDF turtle format, ontology annotation triples that come before the `rdf:type owl:Ontology` triple are lost. Is there supposed to be such a dependence on the order of RDF triples? I will attach an example via pull request.

ansell:
No, it isn't supposed to be order dependent. However, there are some parts of the parser that support streaming, and may have some incorrect assumptions right now as to whether they should delay parsing of triples or parse them when they first receive them.

balhoff:
Thanks, both of you!

PR #575, opened 2016-11-15
Failing test showing dependence on order of RDF triples for ontology annotations.

balhoff:
This illustrates issue #574.

### DIFF
--- a/contract/src/test/java/org/semanticweb/owlapi/api/test/syntax/rdf/TripleOrderingTestCase.java
+++ b/contract/src/test/java/org/semanticweb/owlapi/api/test/syntax/rdf/TripleOrderingTestCase.java
@@ -1,0 +1,27 @@
+package org.semanticweb.owlapi.api.test.syntax.rdf;
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+import org.semanticweb.owlapi.api.test.baseclasses.TestBase;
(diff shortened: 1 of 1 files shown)


ansell:
The place where the fork in behaviour occurs is:

https://github.com/owlcs/owlapi/blob/version4/parsers/src/main/java/org/semanticweb/owlapi/rdf/rdfxml/parser/TripleHandlers.java#L937

Not sure where in the code is the cleanest place to fix it though, AnnotationAssertionAxioms should be checked and migrated into the same Set as other Ontology IRI annotations (and possibly Ontology Version IRI annotations). It could be checked at the end of the parse, or when the OntologyID is changed, or they could be deferred until the end of the parse, but there could also be other simpler places.

ignazio1977:
Stopping the annotation assertion handler from handling this form of annotation in streaming would mean a fairly common form of annotation would need to hang around in memory as RDFTriple for a longer time, so that has potential performance risks.

At the end of parsing, there is currently a phase in which the ontology IRI is picked (it is possible to have more than one IRI with type ontology... well, anyway.); if multiple candidates exist, one that is not a subject of annotation assertions is selected - so if this situation and multiple IRIs marked as ontologies are present, this will affect the result.

I believe the best place to handle this issue is at that point.

ignazio1977:
I've included this test in the fix for #574, cheers

PR #609, opened 2017-01-13
Test demonstrating Manchester expression parser bug with data cardinalities

balhoff:
The Manchester syntax expression parser fails to parse data cardinality expressions without fillers. I.e. it successfully parses something like:

`hasAge exactly 1 xsd:int`

but fails to parse:

`hasAge exactly 1`


ignazio1977:
Nice test, cheers.

PR #631, opened 2017-02-25
Move protected TOP/BOTTOM constants from impl/org.semanticweb.owlapi.reasoner.impl.DefaultNode into  appropriate subclasses.

sesuncedu:
Referencing a subclass from a field  initializer in the superclass can cause deadlocks in class loaders.


ignazio1977:
I think there's a better solution for this in master - I've rejiggled a few tidbits so that there is no need to create an ad hoc owldatafactory to access constants.

ignazio1977:
Does this still apply after the other pull requests?

sesuncedu:
The classloader bug is still there.

ignazio1977:
Merged in 61b97607f8cdb722a6baebb9eeba917a6746dd1b

PR #632, opened 2017-02-25
Rearrangement

sesuncedu:
Applied code rearranger to reformatted code (as separate step). 
Might want this, might not.


sesuncedu:
oops - wrong base

PR #637, opened 2017-02-27
LaTex Rendering Improvements

cogan-shimizu-wsu:
Fixed the formatting of the code. 
Fixed rendering of InverseObjectProperty to properly expand First, Second properties and escape illegal LaTeX Characters and render only ShortForm namespace.


cogan-shimizu-wsu:
The code still has flags for turning "formatter:on" and "formatter:off" which were ignored by the other contributor. 

Should the tags be removed or should other contributor's changes supersede these tags?

ansell:
Opening a new issue would be great for any cases where the formatting is broken. The main issue you are trying to fix is the latex rendering and those changes don't appear to be in this pull request so far.

ansell:
By that I mean that the commits are there, but the diff to the version5 branch only shows whitespace and reformatting changes. The GitHub interface may be broken, or the changes may have been nullified by the merge commit.

ignazio1977:
> Should the tags be removed or should other contributor's changes supersede these tags?

Looks like IntelliJ does not understand the Eclipse tags. I'll revert all those undesired reformats separately.

PR #641, opened 2017-03-10
Add predicates for  "simple literal" so rendering doesn't add explicit xsd:string to everything.

sesuncedu:
This might possibly need to be set by a flag; alternatively, only LiteralImplPlain could admit to being simple.

This almost duplicates isPlainLiteral, except for one rio test case, involving a constructed triple with explicit datatype PlainLiteral.

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
+++ b/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
@@ -161,6 +161,10 @@ public class RDFLiteral extends RDFNode implements org.apache.commons.rdf.api.Li
         return OWL2Datatype.RDF_PLAIN_LITERAL.getIRI().equals(datatype);
     }
 
+    public boolean isSimpleLiteral() {
+        return datatype.equals( OWL2Datatype.RDF_LANG_STRING.getIRI()) ||
+                datatype.equals(OWL2Datatype.XSD_STRING.getIRI());
+    }
     @Override
     public int compareTo(@Nullable RDFNode o) {
         checkNotNull(o);
--- a/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
+++ b/api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
@@ -161,6 +161,12 @@ public class RDFLiteral extends RDFNode implements org.apache.commons.rdf.api.Li
         return OWL2Datatype.RDF_PLAIN_LITERAL.getIRI().equals(datatype);
     }
 
+
+    /**
+     *
+     * @return true if this node is an RDF 1.1 Simple Literal
+     */
+
     public boolean isSimpleLiteral() {
         return datatype.equals( OWL2Datatype.RDF_LANG_STRING.getIRI()) ||
                 datatype.equals(OWL2Datatype.XSD_STRING.getIRI());
--- a/api/src/main/java/org/semanticweb/owlapi/model/OWLLiteral.java
+++ b/api/src/main/java/org/semanticweb/owlapi/model/OWLLiteral.java
@@ -71,6 +71,8 @@ public interface OWLLiteral
         return false;
     }
 
+    default boolean isSimpleLiteral() {return false;}
+
     /**
      * Gets the lexical value of this literal. Note that if the datatype is
      * {@code rdf:PlainLiteral} then the abbreviated lexical form will be
--- a/api/src/main/java/org/semanticweb/owlapi/model/OWLLiteral.java
+++ b/api/src/main/java/org/semanticweb/owlapi/model/OWLLiteral.java
@@ -71,6 +71,10 @@ public interface OWLLiteral
         return false;
     }
 
+    /**
+     *
+     * @return true if this literal is an RDF 1.1 Simple Literal
+     */
     default boolean isSimpleLiteral() {return false;}
 
     /**
--- a/contract/src/test/java/org/semanticweb/owlapi/api/test/literals/LiteralTestCase.java
+++ b/contract/src/test/java/org/semanticweb/owlapi/api/test/literals/LiteralTestCase.java
@@ -64,10 +64,17 @@ public class LiteralTestCase extends TestBase {
     }
 
     @Test
+    public void testStringIsSimple() {
+        OWLLiteral literal = Literal("hello world");
+        assertEquals(literal.getDatatype(),OWL2Datatype.XSD_STRING.getDatatype(df));
+        assertTrue(literal.isSimpleLiteral());
+    }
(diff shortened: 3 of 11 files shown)


sesuncedu:
Just realized there's  missing doc comment on two API methods

sesuncedu:
Also, this can be a breaking change for VCS users, but they would break if they switched from 4 to 5 (which is how I noticed this in the first place).

ignazio1977:
Looks ok to me. I'd wait for @ansell to comment since it's RIO relevant. Can you pick squash and merge so that it's just one commit?

sesuncedu:
I've discovered the joys of the github squash merge (and also the joy/pain of shelving, resetting / force-pushing  personal forked branches, and unshelving. Also shelving + export/import by patch). 
Hypocritically , I haven't checked to see if the API changes need a package minor bump since I used the Maven intellij project, not the gradle one, so I don't have the  tasks set up.  That is not a blocker for this PR :) 
The main possible block is whether this needs an option, and what the options should control. LangString literals always get the plain and simple treatment, but strings can go either way.  It might be possible to track a "was simple literal" method, since it seems that they get converted inside one class, so we could preserve, but since the spec is broken, quod libet. 
I haven't checked to see how reasoners deal with langString literals vs plain literal facets but since most of them didn't do plain literals properly, meh.

ignazio1977:
Reasoners dealing with owlapi 5 deal with them adequately - meaning, they don't fail.
But then I'm only talking about Openllet and Hermit 1.3.8.500.

I.

sesuncedu:
Hermit had a bug in  the langRange matching - they followed the examples in
the PlainLiteral  spec, instead of the text, and the spec copied the wrong
set of examples from the RFC 🤦‍♂️.  I patched it but I don't know if it
made it in to mainline Hermit.

Hermit-why you hide away in svn cave?

ignazio1977:
If you raised the patch against the repo hosted by Phil https://github.com/phillord/hermit-reasoner then likely not - but if you issue it on my fork I'll get it released. My fork is currently used by Protege, so that's likely to land in the face of numerous users.

sesuncedu:
Now that'll be either  an easy, or a nightmare  merge :)

(BTW, whilst I was cleaning up the home linux box, switching to netatalk 3
and getting rid of all the appledoubles, which involved hypnotoad induced
staring at scrolling terminals, I saw a checkout of the owlapi svn
repository scroll by.

api/src/main/java/org/semanticweb/owlapi/io/RDFLiteral.java
```diff
@@ -161,6 +161,16 @@ public boolean isPlainLiteral() {
         return OWL2Datatype.RDF_PLAIN_LITERAL.getIRI().equals(datatype);
     }
 
+
+    /**
+     *
+     * @return true if this node is an RDF 1.1 Simple Literal
```
ansell: I don't think this matches the definition. The RDF-1.1 Concepts and Abstract Syntax paragraph in question has two parts that I think are distinct:

https://www.w3.org/TR/rdf11-concepts/#section-Graph-Literal

The first is:

```
Please note that concrete syntaxes MAY support simple literals consisting of only a lexical form without any datatype IRI or language tag. Simple literals are syntactic sugar for abstract syntax literals with the datatype IRI http://www.w3.org/2001/XMLSchema#string.
```

Then the paragraph continues, but I think it is not referring to "simple literal" for the second part, partly because "similiarly" isn't used to extend definitions, it is just the opposite of "conversely", which is a contrasting/similarity term:

```
Similarly, most concrete syntaxes represent language-tagged strings without the datatype IRI because it always equals http://www.w3.org/1999/02/22-rdf-syntax-ns#langString.
```

I don't mind having a common boolean method here that covers both for consistency, but I don't think we should reuse the informal "simple literal" term for its name.

sesuncedu:
Damn-you're right.

 I was misremembering ; I was thinking that simple literal was was the
umbrella term for both of the forms with no explicit data type. Though
reading the description again,  I may have merged the two in my head.

The SPARQL 1.1 definition is absolutely clear.

Now I can't think of a good name.
 They're not datatypeless (that's the whole bloody point).
They're not PlainLiterals.
They are typelesslyRenderable by renderers that support it, but that's
ugly.

I will go with the last one unless anyone has a better idea. I might also
get rid of of the method on literal, and add a static method in a Render
support class that does the test for langString or string.

(SPARQL grammar offers no help, as it's all crammed into RDFLiteral.

  RDFLiteral ::=   String <https://www.w3.org/TR/sparql11-query/#rString> (
LANGTAG <https://www.w3.org/TR/sparql11-query/#rLANGTAG> | ( '^^' iri
<https://www.w3.org/TR/sparql11-query/#riri> ) )?)

ansell:
".couldHideTypeInDocuments()"?

ignazio1977:
`.isDatatypeOutputOptional()`

Also, any class name, concept, framework, or begowned older man which is described as 'plain', 'simple', or both, ain't. I'm pretty sure someone has also written a code smell detector that pinpoints this sort of weasel names.

sesuncedu:
Whilst trying to find out which datatypes SPARQL 1.1 must recognize, I found an extra-special definition 
>simple literal denotes a plain literal with no language tag.

(§17.1) 

> **Definition: Simple Literal**
>The set of Simple Literals is the set of all RDF Literals with no language tag or datatype IRI.
(which is  by the proper definition in §18.1.2)

It would have been nice if the OWL, SPARQL and RDF working groups could all have synchronized their calendars :-)

sesuncedu:
Before I recode this:  should there be a renderer control flag, and if so, should it default to dropping the xsd:string where possible?
 .sansDataType

ignazio1977:
Not sure if there's any advantage in being able to switch this on/off. I assume 99% of the time it would be left in the default position, and this position is to skip explicit datatypes when unnecessary.

If you fancy keeping the flexibility, sounds like a new entry in `ConfigurationOptions`.

nvbach91:
> If you fancy keeping the flexibility, sounds like a new entry in `ConfigurationOptions`.

Hi @ignazio1977 ,

I looked through `ConfigurationOptions`, `OWLOntologyLoaderConfiguration` and `OWLOntologyWriterConfiguration` and the option to explicitly output the `xsd:string` datatype to e.g. `OWL/XML` is not there yet. 

I assume no one has yet to come to the point where they need this flexibility yet. But I personally do need to keep the explicitness and I'd like to implement this and then create a PR. Can you pls point me to the right direction where do I change the code? Should I modify the `...renderer.java` files or do I just modify the `isSimpleLiteral` method?

ignazio1977:
I would change the renderer classes, as this is a syntax detail.

PR #651, opened 2017-04-01
Update Examples.java

RadW2020:
I think  a DataFactory object is needed here


RadW2020:
mmm sorry I missed that. 
I thought this was an stand alone file like this one [link](https://github.com/RadW2020/owl-api/blob/master/contract/src/test/java/org/coode/owlapi/examples/Examples.java)

ignazio1977:
I'm happy for a change like that to help with readabilty of the example, but the variable should be named to not conflict with the existing one. Otherwise it will just cause Eclipse to annoy me with warnings :-)

RadW2020:
I think it is a good a idea to have a stand-alone example file for those who get first contact with this project. 
If you like this idea I can work on it.

ignazio1977:
Contributions always welcome!

ignazio1977:
Merged

PR #678, opened 2017-06-27
Add Time and Prov vocabulary

AlexTo:
This commit adds W3 Time and Prov vocabulary as specified here

https://www.w3.org/2006/time

https://www.w3.org/ns/prov-o


ignazio1977:
Merged

PR #703, opened 2017-09-17
Fixes #702

### LINKED ISSUE #702: Improve serialization
In our project (https://github.com/JPL-IMCE/gov.nasa.jpl.imce.ontologies.public), we've experienced problems similar to https://douroucouli.wordpress.com/2014/03/30/the-perils-of-managing-owl-in-a-version-control-system/. I was glad to hear about #273. However, with versions 4.2.8 and 5.0.5, there are still problems with serialization variability with annotations and expressions.

I took a look and noticed that `org.semanticweb.owlapi.util.OWLAPIStreamUtils#equalStreams` expects the streams to be sorted; however, this is not generally true.

In several places, there are calls to various sorting methods defined in `org.semanticweb.owlapi.util.CollectionFactory` that effectively mask exceptions
in case the ordering is unstable. Why would the order be unstable?

This led me to look at `OWLObject` which inherits `Comparable<OWLObject>`.
Why would the order be unstable? 

The answer turned out to be conceptually simple: `uk.ac.manchester.cs.owl.owlapi.OWLObjectImpl#compareTo`.

For the ordering to be stable, then `org.semanticweb.owlapi.model.HasComponents#components` must produce an ordered stream.

The simplest way to guarantee this is to ensure that everything is sorted by construction.
For `OWLObject`s corresponding to the OWL2 spec, this means sorting everything at construction.
For `SWRLObject`, the order must be preserved at construction.

A consequence of this is that all the sorting methods on `org.semanticweb.owlapi.util.CollectionFactory` can be finally deleted because the ordering will be stable, by construction.

Of course, the trick is in the details of doing all this...

NicolasRouquette:


NicolasRouquette:
I believe that I've addressed your concerns by removing the extra sorting.
Is there anything else that I can to do facilitate merging this fix?

ignazio1977:
I've been giving this some thought and I think your first implementation (including sorted map pointers and all) should be included as an OWLOntology wrapper, which would enable two features: first, it would enable you - and anyone else spotting sorting bugs in the future - to ensure their desired sorting, and second it would provide a convenient benchmark for order and performance.

I'll fold this into your latest pull request and merge the code as soon as I can complete it.

ignazio1977:
Pull request included in version5, version4 and master for version 6.

Issue #720, opened 2017-12-19
Guice Dependency in Distribution is Wrong


mnmercer:
We recently upgraded OWLAPI in our application from version `5.0.3` to `5.1.3` and are running into issues with the imported guice package in the `distribution` manifest. The `distribution` pom in version `5.0.3` optionally imports `com.google.inject`.
```xml
<Import-Package>
    !sun.misc,
    !com.google.inject.internal.*;resolution:=optional,
    org.openrdf*;version="[4.0,5.0)",
    info.aduna.iteration;version="[4.0,5.0)",
    *
</Import-Package>
```
However the `5.1.3` pom includes the `com.google.inject` package with an incorrect package version number. The exported package version from guice `v4.1.0` is `1.4`, but this pom has `4.1.0`.
````xml
<Import-Package>
    !sun.misc,
    org.eclipse.rdf4j.*;version="[2.0,3.0)",
    com.google.inject.*;version=4.1.0,
    org.apache.commons.io.*;version=2.5,
    org.apache.commons.logging.*;version=1.2,
    *
</Import-Package>
````
The way I see it, there are two options. Either the `com.google.inject` import should be optional like it was before or the package number should be fixed.

**Comment by ignazio1977** [comment]
> The exported package version from guice v4.1.0 is 1.4

I am trying to figure out where the 1.4 comes from, can you clarify how to verify that?

**Comment by ignazio1977** [comment]
> !com.google.inject.internal.*;resolution:=optional,

That's an exclusion for the internal package, not an inclusion

**Comment by skahmann** [comment]
Ah yes, they are a bit different. I would think you could just remove the manual import of `com.google.inject.*;version=4.1.0,` because I would think the * import at the end of the list would catch it. If not, you'll need to use 1.4 as the version. If you look at the guice jar, they use version 1.4 for their exported packages:

```
Export-Package: com.google.inject;version="1.4",com.google.inject.spi;
 version="1.4",com.google.inject.matcher;version="1.4",com.google.inje
 ct.util;version="1.4",com.google.inject.binder;version="1.4",com.goog
 le.inject.name;version="1.4"
```

So it should probably be `com.google.inject.*;version=1.4`. You may actually need `com.google.inject*;version=1.4` (note the missing period) to catch the `com.google.inject` package.

**Comment by skahmann** [comment]
Removing the line entirely yields the following in the imports list:

```
com.google.inject;version="[1.4,2)",com.google.inject.assiste
[...]

PR #722, opened 2017-12-21
Cleaned up Import-Package for distribution

mnmercer:
I removed the `com.google.inject` import and added an exclusion of `javax.annotation` to the pom for `distribution`. Addresses issue #720.


ignazio1977:
Merged in version5

PR #761, opened 2018-05-19
[WIP] Update OWLAPIObo2Owl.java

cmungall:
Do not inject declarations.

This causes problems when converting from obo to owl, when external ontology classes are referenced


ignazio1977:
Merged in 12a2255ad1d55c6fe21dd53b79bd8ef1806255ae

PR #766, opened 2018-06-10
Fixes #765 by updating OBO_IN_OWL capitalization to oboInOwl

### LINKED ISSUE #765: OBOPrefix.OBO_IN_OWL is improperly capitalized
OBOPrefix.OBO_IN_OWL is currently set to "oboInOWL" capitalization, while the correct should be "oboInOwl".

autayeu:


ignazio1977:
This pull request is to fix the 1.2 parser, but I'm not sure if the capitalisation changed between 1.2 and the latest specs. The capitalisation is correct in the newer parser, and the old parser is disabled by default in the latest OWLAPI releases.

PR #804, opened 2018-11-11
Fixed AD again

mrnolte:
I made another mistake when trying to fix the AD resulting in the same behavior as before. This should work now.


ignazio1977:
Merged in 957ae51afa41da8e15a1fd8e557bf4193a163a7b

PR #833, opened 2019-03-22
Update OBOFormatWriter.java to emit OBO 1.4 compliant text

althonos:
Hi there, 

while working on the OBO 1.4 syntax with @cmungall we realised `owlapi` was not emitting proper OBO for property values, since the newer syntax requires the value to be escaped at all times. This PR should take care of that !


cmungall:
did we check this roundtrips?

ignazio1977:
It doesn't - but for some reason the unit tests didn't fail while verifying the pull request.

cmungall:
Looks like this was prematurely merged, sorry!

This is due to the horrific base code which I am responsible for. It should only be the value that is quoted, not the optional xsd datatype

ignazio1977:
Ok, understood, cheers.

Issue #848, opened 2019-05-14
OBOFormatParser does not allow custom tag value pairs


matthewhorridge:
The spec seems to indicate that custom tag value pairs should be processed, but any frame with a custom tag value pair generates a parse error.  This happens in the OBOFormatParser class - see https://github.com/owlcs/owlapi/blob/ba4aede81b92e91bc9d7af73ddb7f9dd3649a687/oboformat/src/main/java/org/obolibrary/oboformat/parser/OBOFormatParser.java#L523

The OBO2OWL translator deals with unknown tag value pairs by translating them into annotation assertions.  

I believe that custom tags, translated as annotation values should be supported.

cc @cmungall

cmungall:
I think in retrospect it was a mistake allowing that. The actual behavior of the parser at the moment is incredibly useful as undeclared tags happen as a result of errors a lot (maybe I hand-edit too much).

@althonos what is the behavior of your parser here?

althonos:
The spec allows custom tag values *in the header only*, so my parser will allow that, but my OWL translator will not process them at all. It will fail on unknown tags in other frames.

ignazio1977:
Looks like there are dissenting opinions :-)

Three possibilities:
- header
- anywhere
- nowhere

Which one do you reckon it's best?

althonos:
I think the header-only was thought as a compatibility option for older OBO files, but in order to translate header clauses to annotation values the property value clause can already be used explicitly. As Chris said, not allowing unreserved clauses helps detecting the one that are typos... but there are probably only a handful of persons that are hand-editing.

matthewhorridge:
Isn't it possible to declare these non-builtin tags before they are used?  I seem to recall isMetaTag or something like that.

cmungall:
> Looks like there are dissenting opinions :-)

I defer to @althonos! Yes, it's quite correct that undeclared tags are only permitted in the header according to the spec.

The current OWLAPI behavior is therefore in fact correct.

However, the spec is underspecified in terms of what happens to these in translation. Currently the tag is translated to an oboInOwl property in the OWLAPI which does not contradict anything in the spec. We can formalize this existing behavior in the spec.

> Isn't it possible to declare these non-builtin tags before they are used? I seem to recall isMetaTag or something like that.

Nah, for declaring a Typedef as an AP

Anyway, I think go ahead and make the release without any changes here. We can leave this open til we properly thrash out the details but no sense holding up the release on this

[...]

PR #850, opened 2019-05-23
Support for custom tags in OBO file parser

fanavarro:
I have added support for custom tags in OBO files (related issue #848). The obo parser has been modified in order to treat as a custom tag any unrecognized tag, considering that the value of any custom tag would be an unquoted string. Also, enumerate comparisons using "==" have been replaced by equals method, which should be more robust.


fanavarro:
@ignazio1977 thanks for your clarification. I usually try to avoid "==" comparisons when equals method is available.

ignazio1977:
Merged in version 5. I'll cherry pick to the other branches

Issue #704, opened 2017-09-20
SyntacticLocalityModuleExtractor and AtomicDecompositionImpl are inconsistent


mrnolte:
Hey there,
I observed an inconsistency between `SyntacticLocalityModuleExtractor ` and `AtomicDecompositionImpl `. Consider the ontology [enm.enanomapper.14](https://github.com/owlcs/owlapi/files/1317985/enm.enanomapper.14.zip), which I downloaded [here](http://doi.org/10.5281/zenodo.439510) and will refer to as _ENM_. ENM imports the ontology [_GO Slim_](http://purl.enanomapper.org/onto/external/go-slim.owl), which contains two axioms _a1_ and _a2_. According to `AtomicDecompositionImpl ` for ENM, they each form single element atoms which depend on a third non-empty atom _c_. When extracting the module of ENM with respect to the signature of GO Slim using `SyntacticLocalityModuleExtractor `, I receive only the axioms a1 and a2. The fact that c is not a subset of the received module violates the logical guarantee given by the atom characteristics. Below is some example code to recreate the error:
```
final OWLOntologyManager manager = OWLManager.createConcurrentOWLOntologyManager();
System.out.println("Loading Ontology");
final OWLOntology ontology = manager.loadOntologyFromOntologyDocument(new File("enm.enanomapper.14.orig"));
final SyntacticLocalityModuleExtractor extractor = new SyntacticLocalityModuleExtractor(manager, ontology, ModuleType.STAR);
System.out.println("Extracting Module");
final Set<OWLAxiom> module = extractor.extract(go.signature().collect(Collectors.toSet())).stream().filter(axiom -> axiom.isLogicalAxiom()).collect(Collectors.toSet());
System.out.println("Module of 'go':");
module.forEach(axiom -> {
	System.out.println("\t" + axiom);
});
System.out.println("Calculating atomic Decomposition");
final AtomicDecomposition atomicDecomposition = new AtomicDecompositionImpl(ontology, ModuleType.STAR);
final Set<Atom> atomicSuperset = new HashSet<>();
ontology.logicalAxioms().forEach(axiom -> atomicSuperset.add(atomicDecomposition.getAtomForAxiom(axiom)));
System.out.println("Atomic overlap of 'go':");
int i = 1;
for (final Atom atom : atomicSuperset) {
	System.out.println("\tAtom " + i++ + ":");
	System.out.println("\t\tAxioms:");
	for (final OWLAxiom axiom : atom.getAxioms()) {
[...]

PR #906, opened 2020-02-20
General atomic decomposition

mrnolte:
Hey there,
For some time I have been wanting to propose new interfaces and a new implementation of  both module extraction and atomic decomposition. The main reasons are:

- Currently there are two implementations of the AD and three of location-based module extraction in the OWL API TOOLs. This is quite a mess and difficult to maintain. For example, if someone just wants to extract a module, they would usually use `uk.ac.manchester.cs.owlapi.modularity.SyntacticLocalityModuleExtractor`, but neither of the two AD implementations use that. Therefore, changes have to be made three times and while some implementation might be correct, another might contain bugs, as in #704 .
- The current implementations are rather clunky and not very readable. Bugs are hard to find.
- Both current AD implementations depend on their respective implementations of Locality based module extraction. There are other module extraction algorithms, though, for instance, Datalog based module extraction (Romero, Kaminski, Grau, Horrocks. "Module extraction in expressive ontology languages via datalog reasoning", 2016). There should be a way to specify the used module extraction algorithm.

Therefore, I refactored a lot, based on Tsarkovs "Improved Algorithms for Module Extraction and Atomic Decomposition", 2012, made the code easier to read and separated the AD from locality based module extraction.

Cheers,
Robin


ignazio1977:
Good work. I've got a few questions, mostly to do with paperwork and to minimise impact on existing users.

- Are you happy for me to add the same licence headers as the rest of OWLAPI?
- I can't tell precisely which files are brand new and should have your copyright mentioned on them and which ones are slight variations of existing ones and should keep their existing copyright notices. It would be useful if you could add that information.
- The rest of OWLAPI uses spaces not tabs :-) I hope it's ok for me to reformat the code.
- The pull request deletes some files or moves them to different packages (git can't always tell the difference). This might have an impact on existing applications. I would like to put back in the deleted files and rewrite them so they are shells for the new classes, and deprecate them so they can be discarded in the next version.

Backwards compatibility can be a millstone around one's neck, but breaking it outside major version changes is even more so. /sadness

mrnolte:
Thanks for the fast reply!

> Are you happy for me to add the same licence headers as the rest of OWLAPI?

Yes please! Didn't think of it but the changes are intended to be open source.

> I can't tell precisely which files are brand new and should have your copyright mentioned on them and which ones are slight variations of existing ones and should keep their existing copyright notices. It would be useful if you could add that information.

Only `org.semanticweb.owlapi.modularity.locality.SyntacticLocalityEvaluator` and `org.semanticweb.owlapi.modularity.locality.SemanticLocalityEvaluator` use code from the `com.clarkparsia.owlapi.modularity.locality` package. The other classes are newly written according to Tsarkov's paper. Note that I left a copyright notice at the top of both classes, but I did a quick re-check and `com.clarkparsia.owlapi.modularity.locality.SyntacticLocalityModuleExtractor` contained it too (wrongly). Since the Javadoc was off, too, I updated the class a few moments ago.

> The rest of OWLAPI uses spaces not tabs :-) I hope it's ok for me to reformat the code.

Noted that one for the future; please refactor as much as you want!

> The pull request deletes some files or moves them to different packages (git can't always tell the difference). This might have an impact on existing applications. I would like to put back in the deleted files and rewrite them so they are shells for the new classes, and deprecate them so they can be discarded in the next version.

That's odd; there should only added classes, but no deletions. I thought about deprecating the old classes, but up to this point, I did only very basic tests and did not compare the performance of my code, yet. Furthermore, `uk.ac.manchester.cs.owlapi.modularity.SyntacticLocalityModuleExtractor` has some additional features that I left out, for instance, enriching the module with associated Annotation- and DeclarationAxioms and sub- and super classes. Therefore, I did not deprecate the the other classes as a part of this pull request. I intend to do some more evaluation, because I have to do some experiments anyway and will give you a quick feedback then.

Btw, I have to reimplement the datalog based module extraction for my master thesis, too, such that they are applicable for the AD. Should I propose them to the OWL API as well, or do you think a new project would be more suitable? I would need an additional dependency for a Datalog reasoner.

Cheers,
Robin

ignazio1977:
A separate project is probably better; a good interface would actually allow shifting some other implementations to separate projects. It's a tricky balance, keeping dependencies low and functionality high.

ignazio1977:
Hum, git is showing me a commit where files are added and removed - the final result being they're unchanged. Probably just a glitch.

mrnolte:
According to github there is nothing removed:
grafik

tools/src/main/java/org/semanticweb/owlapi/modularity/AbstractModuleExtractor.java
```diff
@@ -22,203 +17,77 @@
...
-	 *
-	 * @return The axioms as specified above
-	 */
+	
 	@Override
 	public final @Nonnull Stream<OWLAxiom> axiomBase() {
 		Stream<OWLAxiom> toReturn = axiomBase.stream();
 		toReturn = Stream.concat(tautologies.stream().flatMap(Set::stream), toReturn);
-		return Stream.concat(globals.stream().flatMap(Set::stream), toReturn);
+		return axiomBase.stream();
```
ignazio1977: Was this the intention? globals is no longer involved and the stream creation with tautologies and base axioms is unused

mrnolte:
Oh. Line 51 and 52 are unnecessary, too. In a first draft, the axiomBase did not contain the tautologies and the globals.

ignazio1977:
Merged in https://github.com/owlcs/owlapi/commit/d5284363a41f4ec0be628b195754b0640312f898

PR #929, opened 2020-05-14
Make pruning window size configurable

reality:
Make the window size for fast pruning step in black box explainer configurable. This is especially helpful for very large ontologies, where window sizes of ten may leave you with a 'fast' pruning duration of weeks, even on a very powerful computer. Ideally, you could make a simple heuristic based on the number of axioms in the ontology (for anyone struggling, values of 1000+ for MONDO and ontologies of a similar sizes has helped a lot), but this will do for now.

If this isn't the preferred style for making a parameter, please let me know and I can change it.


ignazio1977:
MONDO has roughly 86000 axioms - would a heuristic like setting the window to the maximum between 1% of the number of axioms and 10 be likely to be good enough? I'd default it to that value if that sounds reasonable, leaving the option of overriding it if needed.

ignazio1977:
Change looks ok to me, I have one minor variation to make and I also want to apply it to the black box debugger, which uses the same pruning window.

ignazio1977:
merged in 410dd059f5d8e94886890c11876b77ba8463ca63 thank you

Issue #451, opened 2015-10-31
explanation tooling to use owlexplanation


ignazio1977:
Replace/retrofit the explanation tooling to use Matthew's
owlexplanation project. Matthew's code improves over the existing code
and is actively maintained.

fcbr:
Is this on the roadmap for OWLAPI still?

PR #932, opened 2020-05-27
issue 451 fix

ayush609:
explanationmethods.java has some methods for explanationapi

for printing use 

public static void printExplanations(Set<Explanation<OWLAxiom>> expl) {
		Iterator<Explanation<OWLAxiom>> i1 = expl.iterator();
	    while(i1.hasNext()) {
	    	Explanation<OWLAxiom> ex = i1.next();//printing the explanation
	    	System.out.println("Explaination "+ex.getAxioms());//printing the explanation
	    	System.out.println();
	    }
	}


dhoom56 (approved):

ignazio1977:
#451 is for refactoring the OWLAPI dependencies and remove explanation code from it, not just to add another dependency. This patch set is only part of the solution, and must happen as part of a major release.

Issue #416, opened 2015-06-18
IRI.create() allows invalid arguments


LorenzBuehmann:
``IRI.create("http://example.com/father#male or http://example.com/father#female")`` 

does not fail, compared to [IRI Validator](http://www.sparql.org/validate/iri?iri=http%3A%2F%2Fexample.com%2Ffather%23male+or+http%3A%2F%2Fexample.com%2Ffather%23female) .

This results in possible OWLClass objects, whose toString() method returns "female" because this is the remainder value of the IRI.

PR #933, opened 2020-05-28
#416 Add checks for IRI

guvi007:
#416 Added checks for IRI using 2 methods, checked if it forms a valid URL and it doesn't contain spaces.
Signed-off-by: Reeshabh <reeshabhkumarranjan@gmail.com>

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/model/IRI.java
+++ b/api/src/main/java/org/semanticweb/owlapi/model/IRI.java
@@ -18,12 +18,14 @@ import static org.semanticweb.owlapi.util.OWLAPIPreconditions.optional;
 import static org.semanticweb.owlapi.util.OWLAPIPreconditions.verifyNotNull;
 
 import java.io.File;
+import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
 import java.net.URL;
 import java.util.Collections;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import javax.annotation.Nullable;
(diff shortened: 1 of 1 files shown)


api/src/main/java/org/semanticweb/owlapi/model/IRI.java
```diff
@@ -81,6 +83,18 @@ protected IRI(URI uri) {
...
     public static IRI create(String str) {
+
+        try {
+            URL url = new URL(str);
+            if (str.contains(" ")) {
+                throw new MalformedURLException();
+            }
+        } catch (MalformedURLException e) {
+            System.err.println("Invalid IRI: " + str);
+            System.exit(1);
```
ignazio1977, line 94: Thanks for your contribution but it cannot be accepted in the current state.

There is a strategic issue and a few smaller problems with this patch:
- Strategically, whether allowing valid IRIs only or not is a behavioural change that might affect a number of existing applications. One could argue that tools that create invalid IRIs have bugs and should be fixed - which is true. But forcing the issue this way is more likely to result in these systems being stuck on old OWLAPI versions, with the fallout from that being users unable to update OWLAPI version if thy also need these systems in the toolchain.
- Local considerations:
 - Creating a URL for each IRI creation will slow down parsing considerably. We've had to remove a lot of code doing this sort of extra validation because the performance penalty is too high.
 - Throwing the malformed url exception without including the URL in the message will make errors harder to debug
 - Throwing and catching an exception right away increases the time required by the code to run, without increasing readability or structure.
 - System.err shouldn't be used directly from the library - this should go to a logger.
 - `System.exit()` will close the virtual machine. This is a bad pattern in a library; any system using it might appear to crash mid operation, without a chance of catching and managing errors and with very little debug information.

Aside from these issues, IRIs are currently created in parsing and rendering code to hold blank node ids; this change would immediately stop our uint tests from passing.

Issue #776, opened 2018-07-29
Allow creation of tautologies for n-ary axioms


ignazio1977:
From ykazakov #306 

For example, we want to provide explanations of entailments in ELK as derivations (sequences of OWL axioms where every next axiom is derived from previous axioms by a simple rule). For this, we may need to use some tautological axioms, like SubClassOf(A A) or EquivalentClasses(ObjectIntersectionOf(A B) ObjectIntersectionOf(B A)) as initial axioms (if one wants to explain a derivation of a tautology, one always need to start with tautologies). The latter axiom would express that the conjunction is a commutative operation. However, OWL API would "swallow" the second argument of EquivalentClasses and our derivation would be broken. It is very difficult to predict when such things happen.

ignazio1977:
Given the way `OWLDataFactoryImpl` and all `OWLObject` implementations currently work, my opinion is that the best way to accommodate this is a specialised provider and a flag passed to the `OWLObject` implementation to skip duplicate removal.

In the context of this requirement, I don't think roundtrip capabilities from an ontology file are required - switching duplicate removal off for general uses would allow roundtrip of these tautologies to file and back but it would be a behaviour change that would affect other applications as well, and would only manifest in those places where arguments get assembled without checking for duplication. This might be very common - hard to tell in advance.

A specialised provider would not affect existing parsers, so the serialization of tautologies will either create axioms with the tautology removed or cause an exception. It could be possible to avoid throwing exceptions in parsing with a property.

Having said that, a property could be enough to allow the functionality without the need for specialised providers. However, this would not allow to fine tune where duplication would be removed and where it would stay - `OWLDataFactoryImpl` is instantiated as a singleton so there would be no context available.

Cosmopal:
> a specialised provider and a flag passed to the OWLObject implementation to skip duplicate removal.

@ignazio1977 I was trying to have a look at how to incorporate this change. Would this mean adding an aditional provider to the list of providers OWLDataFactory implements already? In that case, that new provider would be an extension of DisjointAxiomProvider ?

Cosmopal:
@ignazio1977 
> a specialised provider and a flag passed to the OWLObject implementation to skip duplicate removal.

I was looking at how to go about making this work. Do we need to create a new (specialised provider)? Can't it be done by adding an overload function which takes in a boolean argument to indicate that duplicates need to be kept? 

I am asking this because, as an example, the code creating set from input arguments is in the definition of the interface EquivalentAxiomProvider

[...]

PR #934, opened 2020-05-28
After

Cosmopal:
Fixed issue https://github.com/owlcs/owlapi/issues/776

Only two axioms require duplicates to create tautologies - subclass (SubClass (A, A) and equivalence Equivalence(A, A)). Out of these, equivalence class axioms remove the duplicates from the class Expressions.

In this fix, an overload of the existing functions is added to EquivalentAxiomProvider to enable user to force the provider to ignore duplicates in the input class expressions.

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/model/axiomproviders/EquivalentAxiomProvider.java
+++ b/api/src/main/java/org/semanticweb/owlapi/model/axiomproviders/EquivalentAxiomProvider.java
@@ -43,6 +43,17 @@ public interface EquivalentAxiomProvider {
 
     /**
      * @param classExpressions equivalent classes. Cannot be null or contain nulls.
+     * @param keepDuplicates whether to keep duplicate items
+     * @return an equivalent classes axiom with specified operands and no annotations
+     */
+    default OWLEquivalentClassesAxiom
+    getOWLEquivalentClassesAxiom(Collection<? extends OWLClassExpression> classExpressions,
+                                 boolean keepDuplicates) {
+        return getOWLEquivalentClassesAxiom(classExpressions, Collections.emptySet(), keepDuplicates);
+    }
+
+    /**
+     * @param classExpressions equivalent classes. Cannot be null or contain nulls.
      * @return an equivalent classes axiom with specified operands and no annotations
      */
     default OWLEquivalentClassesAxiom
@@ -58,6 +69,16 @@ public interface EquivalentAxiomProvider {
     OWLEquivalentClassesAxiom getOWLEquivalentClassesAxiom(
         Collection<? extends OWLClassExpression> classExpressions,
         Collection<OWLAnnotation> annotations);
+
+    /**
+     * @param classExpressions equivalent classes. Cannot be null or contain nulls.
+     * @param annotations A set of annotations. Cannot be null or contain nulls.
+     * @param keepDuplicates whether to keep duplicate items
+     * @return an equivalent classes axiom with specified operands and annotations
+     */
+    OWLEquivalentClassesAxiom getOWLEquivalentClassesAxiom(
+            Collection<? extends OWLClassExpression> classExpressions,
+            Collection<OWLAnnotation> annotations, boolean keepDuplicates);
 
     /**
      * @param classExpressions equivalent classes. Cannot be null or contain nulls.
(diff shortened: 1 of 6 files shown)


api/src/main/java/org/semanticweb/owlapi/model/axiomproviders/EquivalentAxiomProvider.java
```diff
@@ -41,6 +41,17 @@
...
     }
 
+    /**
+     * @param classExpressions equivalent classes. Cannot be null or contain nulls.
+     * @param keepDuplicates whether to keep duplicate items
+     * @return an equivalent classes axiom with specified operands and no annotations
+     */
+    default OWLEquivalentClassesAxiom
+    getOWLEquivalentClassesAxiom(Collection<? extends OWLClassExpression> classExpressions,
+                                 boolean keepDuplicates) {
```
ignazio1977, line 51: This is one possible solution to the reasoner need for tautologies, but it requires the reasoner to be changed to call these methods. Ideally, a solution that can be switched on or off outside the reasoner code would be better. Given that data factories do not come with their own configuration, we're probably restricted to system properties (along the lines of `ConfigurationOptions` anyway). I'm not sure which solution is best at this point.

I'm also not totally convinced only two axiom types need this ability; although those are the ones mentioned so far, removing duplicates is currently done to more axiom types than that (subclass, equivalent classes and properties, disjoint classes and properties, disjoint union, off the top of my head). I think it would be easier to have an option to switch off duplicate removal for all axiom types, not just one.

Cosmopal: > (subclass, equivalent classes and properties, disjoint classes and properties, disjoint union, off the top of my head). I think it would be easier to have an option to switch off duplicate removal for all axiom types, not just one.

Yes, that'd be easier! My current understanding is that all these classes ultimately call the sorted function in OWLAPIStreamUtils.java, which does the duplicate removal. A simple solution can be to set a static parameter "keepDuplicates" set on or off in this util class, which will switch on or off the duplicate removal done by this class. In addition to this, the interfaces would need to give up creating sets of the input arguments and use lists instead. 

This would solve both problems - ability to switch on/off the duplicate removal from outside the reasoner and allowing duplicate removal for the classes which use StreamUtils to create the sorted list of input arguments (which seem to be most of them). 

I originally was trying the above but was under the impression we'd need to allow duplicated only to those axioms which actually need duplicates. Rather than disabling duplicate removal for _all_ the classes, as they might actually require removing them. So, I added the duplicate removal as an additional (optional) call in the StreamUtils, and limited its use to specific interface/implementation.


So should I attempt the above mentioned solution?

sszuev: > Yes, that'd be easier! My current understanding is that all these classes ultimately call the sorted function in OWLAPIStreamUtils.java, which does the duplicate removal. A simple solution can be to set a static parameter "keepDuplicates" set on or off in this util class, which will switch on or off the duplicate removal done by this class. In addition to this, the interfaces would need to give up creating sets of the input arguments and use lists instead.

Hello everyone.

The solution with "OWLAPIStreamUtils.java + static keepDuplicate flag" has limited implementation capabilities - it can be implemented only in the OWL-API-impl.
Different implementations, such as ONT-API, don't even use `OWLAPIStreamUtils` and static flags.  
In my opinion, the better solution is to add config option to the `OWLOntologyLoaderConfiguration`&`OntologyConfigurator`. 
These two classes are parts of OWL-API-api (with is a contract that describes behaviour), altough they are concret classes, not interfaces, unfortunately. 
But that means a lot more work

ignazio1977: https://github.com/owlcs/owlapi/commit/b51e6930bfb26f5bfabe0701c2408782385cd894 this should do the job, it allows the behaviour to be defined fr the whole VM and all factories created within it (with no changes to client code) or to create one factory with different behaviour (no impact on other functionalities but it requires changes to the reasoner).

ignazio1977: Update: that's the first commit for version5; for version4, that's all that can be done without changing the API (i.e., OWLDataFactory uses only Sets as inputs, in version4).

ignazio1977:
Merged as mentioned above.

PR #935, opened 2020-05-29
Updating new AtomicDecomposition

mrnolte:
Hey Ignazio, I recently tested the new AD implementation against the [BioPortal snaphsot](https://zenodo.org/record/439510) and observed some room for improvement. Therefore, this pull request

1. fixes a `NoSuchElementException` that occured when an extracted module was equal to the whole ontology,
2. fixes that `getDependency()` and `getDependents()` sometimes returned the atom itself only when it contained more than one axiom (now the atom itself wount be returned at all),
3. updates the JavaDoc according to 2, along with the correction of some documentation that was erroneous or unspecific, and
4. introduces convenience methods for atoms, for example, to check if an atom contains an axiom or to retrieve only the direct dependency relation.

Although I imported the OWL API auto formatter xml for eclipse, somehow the layouting seems to be wrong again, sorry for this inconvenience.

Cheers,
Robin


ignazio1977:
Good stuff, no worries about the formatting. Merged in 4c179959f9791a70b758cdb8f1a04c5fd067cd95

Issue #954, opened 2020-07-15
Ontology not loaded in case of multiple HTTP redirects


ventuc:
When loading an external ontology, OWL API currently follows only one HTTP redirect.

For example it doesn't work for Dublic Core ontology, because to reach http://purl.org/dc/elements/1.1/ the client must follow 2 http redirects:
1. http://purl.org/dc/elements/1.1/ --> HTTP 302, redirects to https://www.dublincore.org/specifications/dublin-core/dcmi-terms/dublin_core_elements#
2. https://www.dublincore.org/specifications/dublin-core/dcmi-terms/dublin_core_elements# --> HTTP 303, redirects to http://www.dublincore.org/specifications/dublin-core/dcmi-terms/dublin_core_elements.rdf
3. http://www.dublincore.org/specifications/dublin-core/dcmi-terms/dublin_core_elements.rdf is the actual Dublin Core ontology

ventuc:
Hi @ignazio1977 , sorry to bother you, but still no news about the release of the fix?

ignazio1977:
I've finished the first javadoc review, should be able to do the release this evening.

ignazio1977:
5.1.16 released, should be on maven central by tomorrow

ventuc:
> 5.1.16 released, should be on maven central by tomorrow

Thanks!

PR #955, opened 2020-07-15
Ability to follow multiple HTTP redirects when loading ontologies

ventuc:
Fix for issue #954

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi/io/DocumentSources.java
+++ b/api/src/main/java/org/semanticweb/owlapi/io/DocumentSources.java
@@ -220,8 +220,10 @@ public class DocumentSources {
 
     protected static URLConnection connect(OWLOntologyLoaderConfiguration config,
         URLConnection conn, int connectionTimeout, String acceptHeaders) throws IOException {
-        if (conn instanceof HttpURLConnection && config.isFollowRedirects()) {
-            // follow redirects to HTTPS
+    	
+    	boolean retry = true;
+    	while (conn instanceof HttpURLConnection && config.isFollowRedirects() && retry) {
+    		// follow redirects to HTTPS
             HttpURLConnection con = (HttpURLConnection) conn;
             con.connect();
             int responseCode = con.getResponseCode();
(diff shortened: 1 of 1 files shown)


lewismc:
Do you have a test case @ventuc ? It would help to prove the case. Offhand, I personally cannot reproduce this use case... BUT, I think I have experienced it in the past! A test would help so much.

ventuc:
Hi @lewismc,

I don't really have a test case. I made a simple test only to reproduce the bug. At the moment, I get it only when loading Dublic Core.

Without this fix, the following test fails because the document retrieved is not the RDF/XML ontology but an HTML document telling to see another URL.


```java
OWLOntologyLoaderConfiguration config = new OWLOntologyLoaderConfiguration();
config.setFollowRedirects(true);

Optional<InputStream> ontologyStream = DocumentSources.getInputStream(IRI.create("http://purl.org/dc/elements/1.1/"), config, "application/rdf+xml");
assertTrue(ontologyStream.isPresent());
StringWriter sw = new StringWriter();
IOUtils.copy(ontologyStream.get(), sw, StandardCharsets.UTF_8);
String ontology = sw.toString();

assertTrue(ontology.startsWith("<?xml"));
```

ignazio1977:
Cheers @ventuc, I'm trying to figure out the implications of the change, looks correct but I haven't wrapped my head around possible consequences of the cycle yet.

ventuc:
Thanks! I hope you can merge this fix as soon as possible, as for now we're stuck for being not able to import Dublin Core...

ignazio1977:
I found a simpler change that passes the test.
The method was meant to call itself with a rebuilt connection if a redirect http status was found; it instead just rebuilt the connection - so, it follows redirects only once, while it should do so multiple times, until a non redirect URL is found.

However, we also need to keep track of visited URLs. Otherwise a scenario where URL A redirects to URL B and URL B redirects to A would lead to an infinite recursion (or an infinite loop depending on implementation.

ventuc:
Ok, good!
So, should I close this pull request? When are you planning to release the fix? Thank you!

ignazio1977:
Clearing out some javadoc, in a few days.

Issue #959, opened 2020-07-19
DO NOT run Javadoc generation as part of install...


lewismc:
Javadoc generation is NOT... NEVER part of a vanilla install.
Make it part of CI and/or QA... not installation. 
PR, coming up...

lewismc:
`mvn javadoc:aggregate`


```
2 errors
61 warnings
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary for OWLAPI 5.1.15:
[INFO]
[INFO] OWLAPI ............................................. FAILURE [ 42.825 s]
[INFO] OWLAPI :: Interfaces ............................... SUCCESS [  0.395 s]
[INFO] OWLAPI :: Default Implementation ................... SUCCESS [  0.004 s]
[INFO] OWLAPI :: Atomic Decomposition and Explanation ..... SUCCESS [  0.002 s]
[INFO] OWLAPI :: Parsers .................................. SUCCESS [  0.004 s]
[INFO] OWLAPI :: OBO Format ............................... SUCCESS [  0.007 s]
[INFO] OWLAPI :: RDF4J Rio ................................ SUCCESS [  0.004 s]
[INFO] OWLAPI :: Binding and Config ....................... SUCCESS [  0.005 s]
[INFO] OWLAPI :: Compatibility ............................ SUCCESS [  0.003 s]
[INFO] OWLAPI :: Tests and Tutorials ...................... SUCCESS [  0.028 s]
[INFO] OWLAPI :: Distribution ............................. SUCCESS [  0.069 s]
[INFO] OWLAPI :: OSGi and Compatibility Distribution ...... SUCCESS [  0.025 s]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  45.492 s
[INFO] Finished at: 2020-07-19T12:41:34-07:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.2.0:aggregate (default-cli) on project owlapi-parent: An error has occurred in Javadoc report generation:
[ERROR] Exit code: 1 - /Users/lmcgibbn/Downloads/owlapi/api/src/main/java/org/semanticweb/owlapi/search/Filters.java:98: error: reference not found
[ERROR]      * @deprecated use {@link org.semanticweb.owlapi.model.OWLOntology#annotationAssertionAxioms()}
[...]

PR #960, opened 2020-07-19
ISSUE-959 DO NOT run Javadoc generation as part of install

lewismc:
This issue allows a local developer (me :) ) to build the project!

Documented in #959

### DIFF
--- a/distribution/pom.xml
+++ b/distribution/pom.xml
@@ -147,9 +147,9 @@
 					</execution>
 				</executions>
 			</plugin>
-			<plugin>
+			<!--plugin>
 				<artifactId>maven-javadoc-plugin</artifactId>
-				<version>2.10.4</version>
+				<version>${maven-javadoc-plugin.version}</version>
 				<configuration>
 					<show>public</show>
 				</configuration>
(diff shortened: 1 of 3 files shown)


lewismc:
I do not like commenting out code... please advise how this should be implemented for this project. Thank you

lewismc:
I literally cannot build the project without this **fix**.

ignazio1977:
Have you tried associating it to the deploy phase rather than the package phase? I'd expect that to do the trick. Otherwise we should remove the phase annotation and play around to make sure the javadoc are generated before the signatures are generated.

-- build systems are the worst except for all the alternatives

lewismc:
> Have you tried associating it to the deploy phase rather than the package phase?

No... I just tried to install!

> I'd expect that to do the trick.


That's not in any README therefore it is not assumed. I would like to try and make this project more developer friendly.

Just to be clear... I cannot even build the project in its current state. 

> -- build systems are the worst except for all the alternatives

Let's agree to disagree... we should at least be able to build the project. Right now I cannot.

ignazio1977:
> Let's agree to disagree

It was a joke... Anyway.

Not sure why the build doesn't work on Java 8, that's what I use to build version 4 and 5.

There's a switch in the main pom. Set <no-javadoc>false</no-javadoc> to <no-javadoc>true</no-javadoc>, it will skip javadoc generation.

ignazio1977:
ff84f153f89619a192c71e888c2114956a1ebaae should resolve this problem. It also moves the version to 5.1.16-SNAPSHOT.

lewismc:
Thanks @ignazio1977

PR #978, opened 2020-10-22
owlapi-contract (v6): fix TestFiles

sszuev:
two minor changes:
- make tests passed on win (method TestFiles#string - replace `\r\n` with `\n`)
- make test data valid (method TestFiles#strings - it didn't work, empty array was returned)


ignazio1977:
Merged in https://github.com/owlcs/owlapi/commit/f427200b6b447b7b7d274ed21ddc4586a687b9e8 after resolving conflicts.
I've left the normalisation out, I believe the problem was in the data, they should all have had `\n` line terminators but a few didn't. I've checked the code out in a Windows 10 VM with autocrlf=false to ensure line endings are not modified by git and it builds successfully.

PR #979, opened 2020-10-22
owlapi-api (v6): changes in model

sszuev:
This commit is only for **v6** since same changes in **v5** may affect existing client code (if someone uses `OWLPropertyDomainAxiom` abstraction directly, then changes in generic types will lead to a compilation error). 
I think it is safe for v6, since no one is using v6 yet.
For v5 there is another PR https://github.com/owlcs/owlapi/pull/980

the patch contains two things: 
- `org.semanticweb.owlapi6.model.OWLPropertyDomain` has been out of use, it is fixed: `OWLPropertyDomainAxiom` has now two generic params
- `org.semanticweb.owlapi6.model.OWLSubAnnotationPropertyOfAxiom` is extending `OWLSubPropertyAxiom` (for reasons of uniformity)

### DIFF
--- a/api/src/main/java/org/semanticweb/owlapi6/model/OWLClassExpression.java
+++ b/api/src/main/java/org/semanticweb/owlapi6/model/OWLClassExpression.java
@@ -22,7 +22,7 @@ import java.util.stream.Stream;
  * @author Matthew Horridge The University Of Manchester Bio-Health Informatics Group
  * @since 2.0.0
  */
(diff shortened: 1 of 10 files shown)


ignazio1977:
Merged under https://github.com/owlcs/owlapi/commit/445bde58cd191a8abbde880487993c3186339f4e with a tweak to remove more asymmetries.

Not sure if the improvement for version5 is worth the possibility of breakages.

sszuev:
Also have some doubts about PR #980  ->ok, then close

PR #980, opened 2020-10-22
owlapi-api (v5): changes in model 

sszuev:
It is for **v5** only, for *v6* there is another PR  https://github.com/owlcs/owlapi/pull/979

the patch contains two things:

- deletion `org.semanticweb.owlapi.model.OWLPropertyDomain`  - it has been out of use since its creation, so deletion is safe; another option would be changing `OWLPropertyDomainAxiom` to use this class (like in #979), but such solution may affect client's code.
- extending `org.semanticweb.owlapi.model.OWLSubAnnotationPropertyOfAxiom` with super-type `OWLSubPropertyAxiom` - it seems expanding this functionality can't damage any code (but may still break functionality, if someone implement custom searching of `OWLSubPropertyAxiom`, for example, but I think it has very low probability)


sszuev:
from PR #979:
> Not sure if the improvement for version5 is worth the possibility of breakages.

close

PR #1003, opened 2021-07-15
Context-aware ontology storer protocol

### LINKED ISSUE #1002: Serialization to Trig and rdf/json should include a named graph.
In the OWLApi, there is a 1-to-many correspondence between 
- an ontology format (i.e. `org.semanticweb.owlapi.model.OWLDocumentFormat`) and 
- an ontology storer that serializes an ontogy in the desired format (i.e. `org.semanticweb.owlapi.util.AbstractOWLStorer`)

Among these storers , the `org.semanticweb.owlapi.rio.RioStorer` handles several RDF formats, including Trig and RDF/JSON where an ontology IRI should also appear as a named graph context for the ontology's contents. To handle this, the `RioStorer` constructor accept a set of RDF resource contexts:

```
    public RioStorer(OWLDocumentFormatFactory ontologyFormat, RDFHandler rioHandler,
        Resource... contexts) 
```

(see https://github.com/owlcs/owlapi/blob/owlapi-parent-5.1.17/rio/src/main/java/org/semanticweb/owlapi/rio/RioStorer.java#L88)

It turns out that this context is currently always empty!

See: https://github.com/owlcs/owlapi/blob/owlapi-parent-5.1.17/rio/src/main/java/org/semanticweb/owlapi/rio/AbstractRioStorerFactory.java#L54

```
    public OWLStorer createStorer() {
        return new RioStorer(getFormatFactory());
    }
```

This affects the ontology manager where the storer is always constructed independently of the ontology to save. 
This means that for formats like trig and rdf/json, there is simply no way using the ontology manager to serialize 
ontologies with a named graph context.

https://github.com/owlcs/owlapi/blob/owlapi-parent-5.1.17/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLOntologyManagerImpl.java#L1322

```
    public void saveOntology(OWLOntology ontology, OWLDocumentFormat ontologyFormat,
        IRI documentIRI) throws OWLOntologyStorageException {
        readLock.lock();
        try {
            for (OWLStorerFactory storerFactory : ontologyStorers) {
                OWLStorer storer = storerFactory.createStorer();
                if (storer.canStoreOntology(ontologyFormat)) {
                    storer.storeOntology(ontology, documentIRI, ontologyFormat);
                    return;
                }
            }
```

Of course, one could create storers outside of the ontology manager; however, it would be better for the ontology manager to create context-aware storers in the first place. This is doable but it requires changing the protocol for creating an `OWLStorer`:

```
public interface OWLStorerFactory extends Serializable, Function<OWLOntology, OWLStorer> {

    /**
     * Create new storer.
     *
     * @return new storer
     */
    OWLStorer createStorer(OWLOntology ontology);
```

This then allows the ontology manager to create an ontology-specific storer like this:

```
    public void saveOntology(OWLOntology ontology, OWLDocumentFormat ontologyFormat,
        IRI documentIRI) throws OWLOntologyStorageException {
        readLock.lock();
        try {
            for (OWLStorerFactory storerFactory : ontologyStorers) {
                OWLStorer storer = storerFactory.createStorer(ontology);
                if (storer.canStoreOntology(ontologyFormat)) {
                    storer.storeOntology(ontology, documentIRI, ontologyFormat);
                    return;
                }
            }
```

This allows context-aware storers like Rio to use the information about the ontology IRI, if available, to set the context:

```
public abstract class AbstractRioStorerFactory extends OWLStorerFactoryImpl {

    protected AbstractRioStorerFactory(OWLDocumentFormatFactory format) {
        super(format);
    }

    @Override
    public OWLStorer createStorer(OWLOntology ontology) {
        return ontology
                .getOntologyID()
                .getOntologyIRI()
                .map(iri -> new RioStorer(getFormatFactory(), SimpleValueFactory.getInstance().createIRI(iri.toString())))
                .orElse(new RioStorer(getFormatFactory()));
    }
}
```

NicolasRouquette:
Fixes #1002

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -6,7 +6,7 @@
 	<parent>
 		<groupId>net.sourceforge.owlapi</groupId>
 		<artifactId>owlapi-parent</artifactId>
-		<version>5.1.17</version>
+		<version>5.1.18</version>
 	</parent>
 	<dependencies>
 		<dependency>
--- a/api/src/main/java/org/semanticweb/owlapi/model/OWLStorerFactory.java
+++ b/api/src/main/java/org/semanticweb/owlapi/model/OWLStorerFactory.java
@@ -13,7 +13,7 @@
 package org.semanticweb.owlapi.model;
 
 import java.io.Serializable;
-import java.util.function.Supplier;
+import java.util.function.Function;
 
 /**
  * Factory for ontology storers. Provide a priority to allow for sorting of ontology storers.
@@ -22,14 +22,14 @@ import java.util.function.Supplier;
  * storer in the second position, it is sufficient for it to pick a priority value strictly between
  * 0 and 1.
  */
-public interface OWLStorerFactory extends Serializable, Supplier<OWLStorer> {
+public interface OWLStorerFactory extends Serializable, Function<OWLOntology, OWLStorer> {
 
     /**
      * Create new storer.
      *
      * @return new storer
      */
-    OWLStorer createStorer();
+    OWLStorer createStorer(OWLOntology ontology);
 
     /**
      * @return format factory for the format parsed by this storer
--- a/api/src/main/java/org/semanticweb/owlapi/util/OWLStorerFactoryImpl.java
+++ b/api/src/main/java/org/semanticweb/owlapi/util/OWLStorerFactoryImpl.java
@@ -13,6 +13,7 @@
 package org.semanticweb.owlapi.util;
 
 import org.semanticweb.owlapi.model.OWLDocumentFormatFactory;
+import org.semanticweb.owlapi.model.OWLOntology;
 import org.semanticweb.owlapi.model.OWLStorer;
 import org.semanticweb.owlapi.model.OWLStorerFactory;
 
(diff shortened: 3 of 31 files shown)


ignazio1977:
I'm no named graph expert, so I have a few basic questions, sorry if the answer is obvious:

- is the named graph IRI always intended to be the ontology IRI? (if the ontology is not anonymous)
- if not, should the value be overridable?
- is this a functionality expected of the library, or would a developer expect to have to take action to make this happen? (i.e., someone using the RIO libraries without OWLAPI in the mix: would they expect to have to set something, or does RIO adopt this behaviour by default, and so OWLAPI is actively preventing an expected behaviour?)

Looking at the implementation, I believe it's easier to add this functionality directly in the RIO storer without changing interfaces, which would reduce the size of the fix considerably.
It's also possible to build a simple override mechanism using the parameter map that exists already on `OWLDocumentFormat` - this could be used if the ontology IRI is not supposed to always be used, or if the choice should be left to the calling application instead of to the OWL API.

I'll put my proposed implementation in a different pull request to simplify comparisons.

NicolasRouquette:
Propose to close in favor of the strategy in #1007

PR #1007, opened 2021-07-24
Fix Trig and rdf/json should include a named graph. #1002

ignazio1977:
My take for #1002 as alternative to #1003

### DIFF
--- a/rio/src/main/java/org/semanticweb/owlapi/rio/RioStorer.java
+++ b/rio/src/main/java/org/semanticweb/owlapi/rio/RioStorer.java
@@ -52,6 +52,7 @@ import javax.annotation.Nullable;
 
 import org.eclipse.rdf4j.OpenRDFUtil;
 import org.eclipse.rdf4j.model.Resource;
+import org.eclipse.rdf4j.model.impl.SimpleValueFactory;
 import org.eclipse.rdf4j.rio.RDFFormat;
 import org.eclipse.rdf4j.rio.RDFHandler;
 import org.eclipse.rdf4j.rio.RDFWriter;
@@ -236,8 +237,8 @@ public class RioStorer extends AbstractOWLStorer {
             // if this is a writer rather than a statement collector, set its config from the format
             // parameters, if any
             addSettingsIfPresent(format);
-            final RioRenderer ren =
-                new RioRenderer(ontology, verifyNotNull(rioHandler), format, contexts);
+            final RioRenderer ren = new RioRenderer(ontology, verifyNotNull(rioHandler), format,
+                contexts(ontology, format));
             ren.render();
         } catch (OWLRuntimeException e) {
             throw new OWLOntologyStorageException(e);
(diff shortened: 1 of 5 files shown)


rio/src/main/java/org/semanticweb/owlapi/rio/RioStorer.java
```diff
@@ -276,14 +277,31 @@ protected void storeOntology(OWLOntology ontology, OutputStream outputStream,
...
             ren.render();
         } catch (OWLRuntimeException e) {
             throw new OWLOntologyStorageException(e);
         }
     }
 
+    private Resource[] contexts(OWLOntology o, OWLDocumentFormat d) {
+        String namedGraph = o.getOntologyID().getOntologyIRI().map(Object::toString).orElse(null);
+        Object namedGraphOverride = d.getParameter("namedgraph", namedGraph);
+        if (namedGraphOverride != null) {
```
NicolasRouquette, line 299: This is an interesting strategy for addressing the issue!

The current behavior is that Trig and rdf/json produce serializations without a named graph. Given that this has been the behavior for a long time now, it seems prudent to me to support it as well as the new behavior. Using a parameter on the ontology format is a neat trick! However, I realize that this will always produce a named graph serialization for Trig and rdf/json (and any other Rio-based format that uses a context). 

What do you think of a 2nd parameter to determine whether to add the `namedGraphOverride` to the contexts?
Let's say that this parameter is called: `serializeNamedGraph` and the value is a boolean.

Then we would have the following possibilities:

- `serializeNamedGraph` = `false` (the default)
   The parameter `namedGraph` is ignored; the API preserves the old behavior.
- `serializeNamedGraph` = `true` (this must be set explicitly)
   The parameter `namedGraph`, if specified, will provide a context if the ontology does not specify an IRI otherwise the ontology IRI will effectively override the value of this parameter.

NicolasRouquette (commented):
Interesting alternative, one that seems less invasive than https://github.com/owlcs/owlapi/pull/1003

rio/src/main/java/org/semanticweb/owlapi/rio/RioStorer.java
ignazio1977: Sounds like you're describing a new parameter in the ConfigurationOptions enumeration.

ADD_NAMED_GRAPH default to false

namedGraphOverride: if present, always use; if absent, check ADD_NAMED_GRAPH: if true, use ontology IRI if available.

This gives us the following scenarios:
- existing code, no changes -> no named graph IRI, i.e., existing behaviour
- set ADD_NAMED_GRAPH to true -> always output ontology IRI as named graph, i.e., same behaviour as your solution
- leave ADD_NAMED_GRAPH false, add namedGraphOverride to specific document format instances: ad hoc output of named graph IRIs, allowing the application to choose what the named graph IRI is (can obviously be set to the ontology IRI)

The advantage in doing it this way is that the behaviour can be switched on for all formats without specific code changes, potentially leaving the choice to third party code, i.e., neither OWLAPI nor calling code need to decide when to output the IRI.

NicolasRouquette: It seems that we are converging on a reasonable strategy. I am not completely sure that I follow your logic above.
Can you update the PR with it?

PR #1044, opened 2022-02-22
Disable the default builds report generation function

CAIWan1998:
That report generation takes time, slowing down the overall build. Reports are definitely useful, but do you need them every time you run the build. We can conditionally disable generating test reports by setting `<disableXmlReport>true<disableXmlReport>`. If you need to generate reports, just add `-DcloseTestReports=false` to the end of mvn build command.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,7 +30,8 @@
 		<no-javadoc>false</no-javadoc>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 		<!-- 5.0.0 is the earliest release on Maven Central for this series -->
-		<last.japicmp.compare.version>5.0.0</last.japicmp.compare.version>		
+		<last.japicmp.compare.version>5.0.0</last.japicmp.compare.version>
+		<closeTestReports>true</closeTestReports>		
 	</properties>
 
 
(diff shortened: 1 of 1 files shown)


mpkorstanje:
This user has been spamming similar MR's across many different projects. Would advise against merging.

ignazio1977:
> This user has been spamming similar MR's across many different projects. Would advise against merging.

158 contributions in two days is certainly not what one regular person does, although the content of this change seems legit (I had a similar change planned at some point).

I'm confused about what the purpose of unleashing so many pull requests is. Hm.

CAIWan1998:
> > This user has been spamming similar MR's across many different projects. Would advise against merging.
> 
> 158 contributions in two days is certainly not what one regular person does, although the content of this change seems legit (I had a similar change planned at some point).
> 
> I'm confused about what the purpose of unleashing so many pull requests is. Hm.

We are graduate students doing research on CI configuration issues and hoping to use Github PR to get some feedback from developers, not out of malice. We are definitely not spamming maliciously

If this PR has affected you, I would like to express my sincere apologies!

ignazio1977:
Nothing to worry about, just curiosity on my part. If you publish a report I'd be curious to read it.

ignazio1977:
About the code, I don't see any dangers in it but:
- just tried `mvn install` with and without report generation, and it takes the same time on my machine
- `<disableXmlReport>` is deprecated starting with version 3.0.0-M6, i.e., the next version up from current

So there's little gain from adding this flag at this time, as it will require more work as soon as I update current build dependencies, while adding more configuration to be aware of when doing builds.

Issue #1056, opened 2022-04-13
unexpected behavior with malformed triples


balhoff:
I'm not sure this is exactly a bug, since the input isn't fully correct OWL. But the behavior is unexpected: if an ontology includes an annotated axiom using a reserved term as the annotated property, an annotation for this axiom is placed on some other completely unrelated axiom. I would instead expect these triples to be left over as "unparsed triples". Here is an example input causing this behavior:

```turtle
@prefix : <http://example.org/> .
@prefix owl: <http://www.w3.org/2002/07/owl#> .
@prefix rdf: <http://www.w3.org/1999/02/22-rdf-syntax-ns#> .
@prefix xml: <http://www.w3.org/XML/1998/namespace> .
@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
@base <http://example.org/> .

<http://example.org/> rdf:type owl:Ontology .

#################################################################
#    Object Properties
#################################################################

###  http://example.org/p
:p rdf:type owl:ObjectProperty ;
   rdfs:subPropertyOf :q .


###  http://example.org/q
:q rdf:type owl:ObjectProperty .


#################################################################
#    Classes
#################################################################

###  http://example.org/A
:A rdf:type owl:Class .


#################################################################
#    Individuals
#################################################################

###  http://example.org/x
:x rdf:type owl:NamedIndividual .


###  Generated by the OWL API (version 4.5.9.2019-02-01T07:24:44Z) https://github.com/owlcs/owlapi


# Here are the triples causing the bad behavior
_:blank1 rdf:type owl:Axiom ;
   owl:annotatedSource _:blank2 ;
   owl:annotatedProperty rdf:first ;
   owl:annotatedTarget :x ;
   rdfs:comment "This rdf:first assertion is annotated" .
```

You can test the result using [ROBOT](http://robot.obolibrary.org) like this:

```
robot convert -i input.ttl -o output.ofn
```

`output.ofn` will look like this:

```turtle
Prefix(:=<http://example.org/>)
Prefix(owl:=<http://www.w3.org/2002/07/owl#>)
Prefix(rdf:=<http://www.w3.org/1999/02/22-rdf-syntax-ns#>)
Prefix(xml:=<http://www.w3.org/XML/1998/namespace>)
Prefix(xsd:=<http://www.w3.org/2001/XMLSchema#>)
[...]

PR #1062, opened 2022-05-19
Avoid dropping axioms or misplacing annotations when malformed axiom annotations are present

balhoff:
This PR addresses #1056. @ignazio1977 could you let me know if you think it's reasonable to address the issue I posted? These changes pass the existing tests, but I found this parser challenging to follow due to the the way it uses side effects.


ignazio1977 (approved):
Looks good. Yeah these parsers are not easy to work with.

PR #1066, opened 2022-06-21
fix xsd name of XSDVocabulary.G_MONTH_DAY

jicro:
The name of ```XSDVocabulary.G_MONTH_DAY``` enum item is fixed, reference page is:  https://www.w3.org/TR/2001/REC-xmlschema-2-20010502/#gMonthDay


ignazio1977:
Fixed in version 4. Keeping open to remember to do the same in version 5 and 6.

ignazio1977:
added to all branches

PR #1073, opened 2022-08-06
Accept JDK 1.9 bytecode from the Jackson dependencies.

gouttegd:
The dependencies from the Jackson project (com.fasterxml.jackson.core:*) contain, in their recent 2.12.6 version, bytecode targeting JDK 1.9.

This cause the build of the OWLAPI :: RDF4J Rio module to fail because we enforce that all our bytecode (including from dependencies) is targeting JDK 1.8 max.

This PR adds the Jackson modules to the list of dependencies excluded from the bytecode check.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -417,6 +417,9 @@
 					                    <exclude>org.opentest4j:opentest4j</exclude>
 					                    <exclude>org.apiguardian:apiguardian-api</exclude>
 					                    <exclude>org.junit.platform:junit-platform-commons</exclude>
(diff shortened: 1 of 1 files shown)


ignazio1977:
Given that I've already included versions of dependencies above 1.9 for security issues and that there's also a request for updating to libraries that require version 11, it's probably best to update the minimum version to 11 and change the version number significantly. I'm going to move to 5.5.0 for Java 11 compatibility and keep version 6 for Java 17

ignazio1977:
Release 5.5.0 will allow Java 11

PR #1074, opened 2022-08-09
Allow unused list triples to be parsed as object property assertions

balhoff:
This supports a kind of unusual use case. This PR enables the RDF parser to convert unconsumed `rdf:first` and `rdf:rest` triples into object property assertions, but only in the case that strict mode is not enabled, and if these predicates have been used in Rbox axioms.

The purpose is to support folks who have RDF datasets which make use of RDF list values, and would like to make use of OWL. One example would be to use property chains to access items in a list, such as inferring a direct relation to each list item. This use case came out of the FHIR RDF working group.

I've tried to ensure that this functionality would only apply when the ontology uses lists this way on purpose; it doesn't strictly adhere to the OWL spec.


balhoff:
The FHIR RDF working group has decided to make some SPARQL updates available to support people using FHIR data with OWL, which transform their RDF lists into another namespace: https://github.com/w3c/hcls-fhir-rdf/pull/111/files

So if this PR is considered too hacky, it could be closed.

PR #1077, opened 2022-11-16
version4: Upgrade RDF4j to 4.2.1 based on what was done for version 5 (commit 5…

sipi:
…5a9f53b71a995b93399a60e19415052da210861)


ignazio1977:
For version 5, this change involved changing the minimum required Java version. Version 4 is used in many systems that are still limited to Java 1.8, this might stop them from being able to get version 4 updates.

sipi:
Indeed, thanks for your answer.

Issue #1083, opened 2023-01-09
OBO serializer incorrectly quoting IRI property values


gouttegd:
The OBO serializer in owlapi 4.5.22 generates slightly incorrect `property_value` tags, in that the value of the property is always quoted, as in the following example:

![](https://user-images.githubusercontent.com/7070631/211374683-56b4fe7f-d5d7-4a66-a274-d4fb0d509730.png)

Per the specification of the OBO Flat File Format (in its 1.4 version), property values should be either _unquoted_ IDs (fully expanded IRIs or CURIEs) or _quoted_ literal values followed by a type specification:

```
property_value-Tag Relation-ID ( QuotedString XSD-Type | ID )
```

So, when the value of a property is an ID, it should _not_ be quoted.

For reference, related ticket on the ODK repository: INCATools/ontology-development-kit#770

gouttegd:
AFAICT, the cause for the systematic quoting behaviour is found in two places:

* The `writePropertyValue` method in [org.obolibrary.oboformat.writer.OBOFormatWriter](https://github.com/owlcs/owlapi/blob/07c9b2ebf82d2a829a28c9ce18c10b79fde60945/oboformat/src/main/java/org/obolibrary/oboformat/writer/OBOFormatWriter.java#L540). This method unconditionally writes the property value in double quotes, regardless of the type of the value.

* The `trGenericPropertyValue` method in [org.obolibrary.obo2owl.OWLAPIOwl2Obo](https://github.com/owlcs/owlapi/blob/07c9b2ebf82d2a829a28c9ce18c10b79fde60945/oboformat/src/main/java/org/obolibrary/obo2owl/OWLAPIOwl2Obo.java#L1078) systematically converts the property value to a Java String (if the value was an IRI, it does that by calling the `getIdentifier` helper method, among other things to convert an IRI into a CURIE).

The result is that by the point we could decide whether the value needs quoting or not (in `OBOFormatWriter.writePropertyValue`), the information that the value is an ID is lost – all we have is a `java.lang.String`.

I can think of three ways of fixing that:

* We make `writePropertyValue` check what the value _looks like_ (does it look like an IRI/a CURIE) and based on that decide whether to quote or not.
* We somehow convey to `writePropertyValue` the information that the value is an ID and therefore should not be quoted (not sure about the best way to convey that bit).
[...]

PR #1085, opened 2023-01-13
OBO serializer incorrectly quoting IRI property values

hkir-dev:
Fix for the issue #1083 

In relation with the OBO Specification of the `property_value`, this fix serializes IRI values without the double quotes.

>property_value
    This tag binds a property to a value in this instance. The value of this tag is a relationship type id, a space, and a value specifier. The value specifier may have one of two forms; in the first form, it is just the id of some other instance, relationship type or term. In the second form, the value is given by a quoted string, a space, and datatype identifier. See 
[IDs](https://owlcollab.github.io/oboformat/doc/GO.format.obo-1_4.html#S.1.6) for more information on legal datatype identifiers.

In order to distinguish IRI values from the quoted string values, heuristic approach applied in the [OWLAPIObo2Owl](https://github.com/owlcs/owlapi/blob/version5/oboformat/src/main/java/org/obolibrary/obo2owl/OWLAPIObo2Owl.java#L1142) is replicated:
```
If Clause has 2 values then it is `relationship type id` + `IRI`. 
If Clause has 3 values then it is `relationship type id` + `value in quoted strings` + `datatype identifier`
```

### DIFF
--- a/contract/src/test/java/org/obolibrary/oboformat/OBOFormatWriterTestCase.java
+++ b/contract/src/test/java/org/obolibrary/oboformat/OBOFormatWriterTestCase.java
@@ -1,25 +1,33 @@
 package org.obolibrary.oboformat;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
+import java.io.File;
 import java.io.FileInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.StringWriter;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
+import java.util.stream.Stream;
 
 import org.junit.jupiter.api.Test;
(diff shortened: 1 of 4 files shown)


matentzn:
@ignazio1977 Happy New Year! :) Hope all is well! 

No pressure (this is our problem not yours), but this issue is a super high priority issue for us. Not just this PR, but also getting it right this time, because we cannot delay our migration to the new Protege any longer..

Do you think you will find time to handle this PR in the next week? If not, can you advice us how/if it _can_ be handled somehow without you?

ignazio1977:
@matentzn happy new year to you too. I'll sort this out today.

ignazio1977:
@matentzn @hkir-dev this pull request was against version 5 of OWLAPI, which will be going out with Java 11 support and is not the one Protege uses (far as I'm aware), I imagine you need the change cherry picked to the version4 branch? I'll do that anyway, just making sure of which version you need the fix in.

hkir-dev:
@ignazio1977 the new Protege release will be using OWLAPI 4.5.22. If you are overloaded, I can quickly create a PR on that branch as well.

ignazio1977:
@hkir-dev no problem, I've done the cherry pick and I'm building 4.5.23 now

ignazio1977:
4.5.23 and 5.5.0 have been released, just waiting for sonatype to sync with maven central

matentzn:
This is absolutely amazing, thank you both for that insanely fast turnaround!!

PR #1091, opened 2023-02-12
OWL to OBO conversion error log

hkir-dev:
Fix for the issue on https://github.com/ontodev/robot/issues/1088 

When processing OWL constructs outside the OBO spec, a nasty error is generated which pollutes the Log files. This error does not affect the conversion, it just logs an error and completes the conversion successfully.

Analysis: 
This error is introduced with this commit: https://github.com/owlcs/owlapi/commit/49eaf8d3e98052e4632b6407b336956c3107ee94
> Fix is to use an injector to get a new manager to create the ontology. A manager provider would be a better solution but there are static methods and interface changes involved.

However, while injecting a new `OWLOntologyManager`, this injector was unable to initialize the `OWLOntologyBuilder` and log an error. But in the consecutive steps, ontology factories are manually set from the source manager, so execution continues successfully.

This fix introduces a dummy `OWLOntologyBuilder` to the injector. By this way the `OWLOntologyManager` is created without error and then the dummy builder is replaced with a proper one as before.

### DIFF
--- a/oboformat/src/main/java/org/obolibrary/obo2owl/OwlStringTools.java
+++ b/oboformat/src/main/java/org/obolibrary/obo2owl/OwlStringTools.java
@@ -19,8 +19,10 @@ import org.semanticweb.owlapi.io.StringDocumentSource;
 import org.semanticweb.owlapi.model.IRI;
 import org.semanticweb.owlapi.model.OWLAxiom;
 import org.semanticweb.owlapi.model.OWLOntology;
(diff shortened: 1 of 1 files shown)


matentzn:
Excellent @hkir-dev, lets see if @ignazio1977 agrees this is the right way to fix https://github.com/ontodev/robot/issues/1088.

@ignazio1977 don't make a new release yet after merging this, there is another major bug we do not have a a strategy for yet: https://github.com/ontodev/robot/issues/1089

PR #1092, opened 2023-02-13
OBO serialisation illegal axiom annotation

hkir-dev:
Fix for the issue: https://github.com/ontodev/robot/issues/1089

OBO serializer is adding a weird axiom annotation when an illegal annotation property is declared:

```
<owl:AnnotationProperty rdf:about="http://www.w3.org/1999/02/22-rdf-syntax-ns#type"/>
```

This fix is one of the two alternative solutions. Other alternative is: https://github.com/owlcs/owlapi/pull/1093

During OWL to OBO conversion, illegal annotation property and its value are filtered.


matentzn:
Alright, making an executive decision now. #1093 wins due to https://github.com/ontodev/robot/issues/1089#issuecomment-1428332507

PR #1093, opened 2023-02-13
OBO serialisation illegal axiom annotation

hkir-dev:
Fix for the issue: https://github.com/ontodev/robot/issues/1089

OBO serializer is adding a weird axiom annotation when an illegal annotation property is declared:

```
<owl:AnnotationProperty rdf:about="http://www.w3.org/1999/02/22-rdf-syntax-ns#type"/>
```
This fix is one of the two alternative solutions. Other alternative is: https://github.com/owlcs/owlapi/pull/1092

During OWL to OBO conversion, illegal annotation property and its value are filtered.

### DIFF
--- a/contract/src/test/java/org/obolibrary/oboformat/Owl2OboTestCase.java
+++ b/contract/src/test/java/org/obolibrary/oboformat/Owl2OboTestCase.java
@@ -183,19 +183,20 @@ class Owl2OboTestCase extends OboFormatTestBasics {
         assertEquals(COMMENT, comment.get().getLiteral());
     }
     
-    @Test
-    void testIllegalOWLAxiomConversion() throws IOException {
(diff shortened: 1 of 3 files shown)


matentzn:
@ignazio1977 this is ready I think. It looks good to my eyes, but maybe sanity check the code.

Will PRs like this made on one version (like 4) be automatically perpetuated to 5 and 6?

Once this is merged, could you do another release of the OWL API? 🙏

ignazio1977:
> Will PRs like this made on one version (like 4) be automatically perpetuated to 5 and 6?

Not automatically but I'll make sure the same changes are applied to version 5 and 6

ignazio1977:
4.5.25 released

matentzn:
AMAZING Thank you!

PR #1120, opened 2024-01-17
remove unused library

mosfet80:
sonatype references are removed


ignazio1977:
That reference allows me to do maven central releases

PR #1121, opened 2024-01-18
Update pom.xml

mosfet80:
Updated owlapi-parent


ignazio1977:
Are you just spamming projects, @mosfet80 ?

PR #1122, opened 2024-01-26
updated libraries

mosfet80:
updated caffeine to version 3.1.8


ignazio1977:
You are not helping. This caffeine version doesn't work with java 8, as you'd know if you built it.

Pass a message to whoever is running this botlike account: this doesn't help.

Issue #1126, opened 2024-02-22
Uncaught NumberFormatException when comparing (large) integers


gouttegd:
This issue was [initially reported](https://github.com/protegeproject/protege/issues/1191) on the Protégé bug tracker.

Protégé encounters an uncaught `NumberFormatException` in the OWL API when trying to display the `http://www.semanticweb.org/ReproduceError/Testing/SuperClass1#subclass` class from the ontology in the attached file: 
[ReproduceErrorNumberFormatException2.txt](https://github.com/owlcs/owlapi/files/14379067/ReproduceErrorNumberFormatException2.txt)

This is because of the large integer value 92740100783 in one of the class expressions defining the class.

In code, the easiest way to reproduce the bug is (assuming `df` is an `OWLDataFactory`):

```java
OWLLiteral smallInteger = df.getOWLLiteral(123);
OWLLiteral largeInteger = df.getOWLLiteral("92740100783", df.getIntegerOWLDatatype());
smallInteger.compareTo(largeInteger);
```

The `compareTo` call raises a `NumberFormatException`. I believe the issues lies in the [equals(OWLObject) method](https://github.com/owlcs/owlapi/blob/c5c2bd29b8d4be6779b8993110f8e7e56af0cf46/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplInteger.java#L158) of the `OWLLiteralImplInteger` class:

```java
public boolean equals(Object obj) {
    if (this == obj) {
        return true;
    }
    if (!super.equals(obj)) {
        return false;
    }
    if (obj instanceof OWLLiteralImplInteger) {
        OWLLiteralImplInteger other = (OWLLiteralImplInteger) obj;
        return literal == other.literal;
    }
    if (obj instanceof OWLLiteral) {
        return ((OWLLiteral) obj).isInteger()
            && ((OWLLiteral) obj).getLiteral().charAt(0) != '0'
            && literal == ((OWLLiteral) obj).parseInteger();
    }
    return false;
}
```

because the call to `parseInteger()` is not protected. If `obj` happens to be an instance of `OWLLiteralImplNoCompression` containing the string representation of a large integer, that call will raise the `NumberFormatException`.

ignazio1977:
Looks like there's also some funny business in the literal parsing, as the value for owl:hasValue does not have a specified datatype and so shouldn't have been parsed to an integer literal.

PR #1127, opened 2024-02-22
Handle possible exception when comparing integer literals.

gouttegd:
This PR is a putative fix for #1126, where a `NumberFormatException` can be raised when trying to compare integer values if one of the values is too large fit into a Java integer.

The fix is simply to catch the exception and fall back to comparing the integers based on their string literal representations.

### DIFF
--- a/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplInteger.java
+++ b/impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplInteger.java
@@ -167,9 +167,14 @@ public class OWLLiteralImplInteger extends OWLObjectImplWithoutEntityAndAnonCach
             return literal == other.literal;
         }
         if (obj instanceof OWLLiteral) {
-            return ((OWLLiteral) obj).isInteger()
-                && ((OWLLiteral) obj).getLiteral().charAt(0) != '0'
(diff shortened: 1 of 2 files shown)


impl/src/main/java/uk/ac/manchester/cs/owl/owlapi/OWLLiteralImplInteger.java
```diff
@@ -167,9 +167,14 @@ public boolean equals(Object obj) {
...
         if (obj instanceof OWLLiteral) {
-            return ((OWLLiteral) obj).isInteger()
-                && ((OWLLiteral) obj).getLiteral().charAt(0) != '0'
-                && literal == ((OWLLiteral) obj).parseInteger();
+            OWLLiteral other = (OWLLiteral) obj;
+            if (other.isInteger() && other.getLiteral().charAt(0) != '0') {
+                try {
+                    return literal == other.parseInteger();
+                } catch (NumberFormatException e) {
+                    return getLiteral().equals(other.getLiteral());
```
ignazio1977, line 175: I'd prefer to not have to rely on exceptions for the control flow, but to do so requires changes to keep track of which integer literals were created with a value within Java integer range, which comes with its own problems.

compareTo() has the same problem, I'll extend the patch set to fix both.

ignazio1977:
merged in version4

PR #1129, opened 2024-03-12
Fixed ontology iri parsing (version5) when annotation imports same iri

vChavezB:
First of all thanks for merging the fix I provided in https://github.com/owlcs/owlapi/pull/1117.

I have also added this same fix for version 5.


ignazio1977:
included in all branches

Issue #1164, opened 2025-03-27
Support anonymous individuals in SWRL rules


phillord:
I am struggling to parse an ontology which was produced by the OWL.

Afraid the code that I am using for parsing is in clojure, but hopefully it should mostly make sense to you. The same code is successfully parsing all my other test cases (https://github.com/phillord/horned-owl/tree/feature/reparse-owl-api/src/ont/owl-xml).

Any thoughts?

````
(try
  (let [documentsource (org.semanticweb.owlapi.io.FileDocumentSource. parse-file)
        config (.get (org.semanticweb.owlapi.OWLAPIConfigProvider.))
        config (.setMissingImportHandlingStrategy config org.semanticweb.owlapi.model.MissingImportHandlingStrategy/SILENT)
        ontology
        (.createOntology
         (org.semanticweb.owlapi.apibinding.OWLManager/createOWLOntologyManager))
        ;; This is the format specific stuff
        parser (org.semanticweb.owlapi.owlxml.parser.OWLXMLParser.)
        ]
    (.parse parser documentsource ontology config))
  (catch Exception e
    (println "Exit with error" resource-file e)
    (System/exit -1)))

(println "Exit with success:" resource-file)
````

 
````
<?xml version="1.0"?>
<Ontology xmlns="http://www.w3.org/2002/07/owl#"
     xml:base="http://www.example.com/iri"
     xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
     xmlns:xml="http://www.w3.org/XML/1998/namespace"
     xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
     xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
     ontologyIRI="http://www.example.com/iri"
     versionIRI="http://www.example.com/viri">
    <Prefix name="o" IRI="http://www.example.com/iri#"/>
    <Prefix name="owl" IRI="http://www.w3.org/2002/07/owl#"/>
    <Prefix name="rdf" IRI="http://www.w3.org/1999/02/22-rdf-syntax-ns#"/>
    <Prefix name="xml" IRI="http://www.w3.org/XML/1998/namespace"/>
    <Prefix name="xsd" IRI="http://www.w3.org/2001/XMLSchema#"/>
    <Prefix name="rdfs" IRI="http://www.w3.org/2000/01/rdf-schema#"/>
    <Declaration>
        <Class IRI="#A"/>
    </Declaration>
[...]

PR #1165, opened 2025-03-28
Fix/anonymous individual swrl rule

phillord:
Addresses: #1164

Need to add a fix and check version 5!


phillord:
Think this is ready to go now.

I guess it needs merging into other branches. Do I need to do version5 and version6?

ignazio1977:
Merged in all branches

